PR #110, opened 2026-04-08
add ROCm support for Windows x86_64, Linux aarch64

ericcurtin:
Extend the existing ROCm plugin system beyond Linux x86_64 to cover all architectures and OSes where ROCm/HIP is available:

- backend.rs: add BackendKind::Rocm to the Windows x86_64 module and probe inferrs_backend_rocm.dll via LoadLibrary, matching the Linux dlopen path; remove the Linux-only cfg guard on the Rocm match arm in main.rs so it compiles on both platforms
- release.yml: build inferrs-backend-rocm on Linux aarch64 (ROCm 6.x sbsa packages via amdgpu-install) and include libinferrs_backend_rocm.so in the aarch64 tarball
- release.yml: install AMD HIP SDK via Chocolatey on Windows x86_64, build inferrs-backend-rocm.dll, and ship it in the Windows zip

macOS, Windows aarch64, and Android are left unchanged — ROCm has no support on those platforms.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -134,7 +134,7 @@ jobs:
           name: inferrs-x86_64-unknown-linux-gnu
           path: inferrs-x86_64-unknown-linux-gnu.tar.gz
 
-  # ── Linux aarch64: main binary + CUDA/Vulkan plugins ─────────────────────
+  # ── Linux aarch64: main binary + CUDA/ROCm/Vulkan plugins ───────────────
   build-linux-arm64:
     name: Build (aarch64-unknown-linux-gnu)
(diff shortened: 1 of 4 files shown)


ericcurtin:
Fixed in 79ca2b8: the Windows ROCm build step now discovers the installed path dynamically by listing `C:\Program Files\AMD\ROCm\` and picking the highest-versioned subdirectory, so it works regardless of which SDK version Chocolatey installs. The Vulkan-not-built-on-Windows observation was also addressed with an explanatory comment — the probe unconditionally returns 1 on non-Linux, so shipping the DLL would be inert.

PR #136, opened 2026-04-09
Fix/think filter cli arg

ilopezluna:
`--think-filter=false` panics at runtime with clap 4.6's debug assertion: *"Arg::is_takes_value_set is required when Arg::is_require_equals_set is set"*.

The fix switches from `default_value_t = true` to `default_value = "true"` with explicit `num_args(1)`, which tells clap the flag always takes exactly one value after `=`.

Includes unit tests covering `--think-filter=false`, `--think-filter=true`, and the default (omitted → `true`).


ericcurtin:
I removed that feature @ilopezluna it was a bad one

ericcurtin:
think_filter no longer exists

PR #142, opened 2026-04-09
Fix: Honor `--max-seq-len` parameter and resolve contiguous tensor warning

co-esquie:
This PR fixes the `--max-seq-len` CLI parameter which was previously defined but ignored. While implementing and testing this feature, we discovered and fixed a related bug where non-contiguous tensors caused warnings during KV cache operations.

### 1. Add `--max-seq-len` support to `run` command (`inferrs/src/run.rs`)

- Added `max_seq_len: usize` field to `RunArgs` with `#[arg(long, default_value_t = 0)]`
- Updated `to_serve_args()` to propagate the value instead of hardcoding `0`

### 2. Use CLI value in engine (`inferrs/src/engine.rs`)

- Modified `load_engine()` to use `args.max_seq_len` when > 0, otherwise fall back to model's default
- This makes the parameter actually functional for both `serve` and `run` commands

### 3. Fix "slice-set only supports contiguous tensors" warning (`inferrs/src/turbo_quant.rs`)

**Problem discovered during testing:** When using `--max-seq-len` with values that trigger KV cache buffer reallocation, the engine would log:
```
WARN inferrs::engine: Engine warm-up failed (non-fatal): slice-set only supports contiguous tensors
```

**Root cause:** The `TurboQuantKvCache` was passing non-contiguous tensor views (from `narrow()` operations) directly to `slice_set()`, which requires contiguous tensors.

**Fix:** Added `.contiguous()?` calls before `slice_set()` at 3 locations:
- Lines 559-562: When copying existing data during warmup buffer growth
- Lines 772-773: When copying existing data during KV buffer reallocation
- Lines 783-784: When writing newly dequantized delta tokens

### 4. Add regression test (`inferrs/src/turbo_quant.rs`)

Added `buffer_realloc_with_narrow_non_contiguous` test that:
- Creates a cache with warmup enabled (exercises the problematic code path)
- Appends 300 tokens to trigger multiple buffer reallocations
- Verifies no "slice-set only supports contiguous tensors" error occurs

## Usage

```bash
# Use model's default sequence length (backward compatible)
./inferrs serve Qwen/Qwen3-8B --turbo-quant=6

# Override with custom context size
./inferrs serve Qwen/Qwen3-8B --turbo-quant=6 --max-seq-len 4096

# Works with run command too
./inferrs run Qwen/Qwen3-8B --max-seq-len 2048 "Hello"
```

## Backward Compatibility

- Default value `0` preserves existing behavior (uses model's default)
- No breaking changes to API or CLI
- All existing tests pass

## Testing

```bash
# Build
cargo build -p inferrs

# Run new test specifically
cargo test -p inferrs buffer_realloc_with_narrow_non_contiguous

# Run all turbo_quant tests
cargo test -p inferrs turbo_quant
```

All 14 turbo_quant tests pass, including the new regression test.

### The `slice_set` issue

`slice_set` is an in-place tensor operation that requires contiguous memory layout. When we call `narrow()` on a tensor, it creates a view with modified strides, which may be non-contiguous depending on the dimension and offset.

Example problematic pattern (before fix):
```rust
let k_valid = kb_old.narrow(2, 0, self.warmup_kv_buf_len)?;  // Non-contiguous view!
new_k_buf.slice_set(&k_valid, 2, 0)?;  // Fails with "requires contiguous"
```

Fixed pattern:
```rust
let k_valid = kb_old.narrow(2, 0, self.warmup_kv_buf_len)?.contiguous()?;  // Now contiguous
new_k_buf.slice_set(&k_valid, 2, 0)?;  // Works!
```

Note: `.contiguous()` is a no-op when the tensor is already contiguous, so there's no performance penalty in the common case.

### DIFF
--- a/inferrs/src/engine.rs
+++ b/inferrs/src/engine.rs
@@ -136,7 +136,11 @@ pub fn load_engine(args: &ServeArgs) -> Result<EngineContext> {
     let arch = raw_config.detect_architecture()?;
     tracing::info!("Detected architecture: {:?}", arch);
 
-    let max_seq_len = raw_config.effective_max_seq_len(&arch);
+    let max_seq_len = if args.max_seq_len > 0 {
+        args.max_seq_len
+    } else {
+        raw_config.effective_max_seq_len(&arch)
+    };
     if max_seq_len < usize::MAX {
         tracing::info!("Model KV cache capacity: {} tokens", max_seq_len);
     }
--- a/inferrs/src/run.rs
+++ b/inferrs/src/run.rs
@@ -94,6 +94,10 @@ pub struct RunArgs {
     /// Path to a WAV audio file to attach to the prompt (Gemma 4 audio models).
     #[arg(long)]
     pub audio: Option<std::path::PathBuf>,
+
+    /// Maximum sequence length (0 = model default)
+    #[arg(long, default_value_t = 0)]
+    pub max_seq_len: usize,
 }
 
 impl RunArgs {
@@ -102,7 +106,7 @@ impl RunArgs {
             model: Some(self.model.clone()),
             revision: self.revision.clone(),
             dtype: self.dtype.clone(),
-            max_seq_len: 0,
+            max_seq_len: self.max_seq_len,
             device: self.device.clone(),
             host: "0.0.0.0".to_string(),
             port: Some(8080),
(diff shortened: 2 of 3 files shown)


ericcurtin (approved):

ericcurtin:
Thanks for the PR, merging pending a green build, welcome to the project

ericcurtin:
If you run "make all" and resubmit, it might fix the fmt issue.

co-esquie:
> If you run "make all" and resubmit, it might fix the fmt issue.

Done, FMT should be applied now

ericcurtin (approved):

PR #147, opened 2026-04-10
fix: statically link cudart so binary has no CUDA DT_NEEDED entries

igorls:
## Summary

Finishes the job started by `646375b` (float8 → 0.7.0) so the inferrs binary really does have **zero CUDA DT_NEEDED entries** — the last piece was `candle-kernels/build.rs:50`, which hard-linked `libcudart.so` for the `libmoe.a` static archive.

This also **unblocks `main` CI on `aarch64-unknown-linux-gnu`**, which is currently red with `/usr/bin/ld: cannot find -lcudart: No such file or directory` — the sbsa `cuda-toolkit-12-6` package does not expose `libcudart.so` at the path the linker searches, but it does ship `libcudart_static.a`.

> Note for reviewers: this PR was originally a float8 0.6.1 vendored fork. Eric superseded the cudarc/float8 half of the problem in `646375b`, so after rebasing onto `main` I pivoted the PR to the actual remaining hard-link. The vendored `float8/` directory is gone; this is now a 2-file change.

## What changed

**`candle-kernels/build.rs`** — swap `dylib=cudart` for `static=cudart_static`:

- `libcudart_static.a` resolves the CUDA driver API (`libcuda.so`) via `dlopen`/`dlsym` internally — verified with `nm -u`, the only undefined symbols are `dlopen`/`dlsym`/`dlvsym`/`dlmopen`/`dlclose` + standard libc. No `cuInit` / `cuMem*` / `cuCtx*` / etc. means no hidden driver link sneaks back in.
- Ships with every `cuda-toolkit-12-6` arch (x86_64, sbsa, Windows), so the aarch64 build-and-link step just works without CI env-var hacks.
- Costs ~500KB after dead-stripping. Fair trade for true "one binary for all CUDA versions" portability.
- Explicit `rustc-link-search=native=\$CUDA_PATH/lib64` (or `lib/x64` on MSVC) so the archive is discoverable without relying on `LIBRARY_PATH`/`LIB`.

**`.github/workflows/ci.yml`** — remove the `LIBRARY_PATH=/usr/local/cuda/lib64` and Windows `LIB=…\lib\x64` workarounds added in `646375b`. They existed only to let the dynamic `-lcudart` resolve; with static linking the build.rs emits its own search path and the env vars are dead weight.

## Verification

Built \`release\` locally on x86_64 Linux with CUDA 13.2:

\`\`\`
$ readelf -d target/release/inferrs | grep NEEDED
    libssl.so.3
    libcrypto.so.3
    libgcc_s.so.1
    libm.so.6
    libc.so.6

$ readelf -d target/release/libinferrs_backend_cuda.so | grep NEEDED
    libgcc_s.so.1
    libc.so.6
    ld-linux-x86-64.so.2
\`\`\`

No \`libcuda\` / \`libcudart\` / \`libcublas\` / \`libcurand\` in either the main binary or the cdylib plugin. The binary can now be shipped once and will dlopen whichever CUDA libs are present at runtime — exactly the \`brew install inferrs\` → "CUDA, ROCm, Metal, Hexagon, OpenVino, MUSA, CANN, Vulkan, CPU" fallback chain target.

\`inferrs --help\` and \`inferrs list\` both run cleanly on the built binary.

## Test plan

- [x] Release build succeeds on x86_64 Linux
- [x] \`readelf -d\` shows no CUDA libs in \`DT_NEEDED\` on \`target/release/inferrs\`
- [x] \`readelf -d\` shows no CUDA libs in \`DT_NEEDED\` on \`target/release/libinferrs_backend_cuda.so\`
- [x] Built binary runs (\`--help\`, \`list\`)
- [ ] CI green on \`aarch64-unknown-linux-gnu\` (was red on \`main\`)
- [ ] CI green on \`x86_64-unknown-linux-gnu\`
- [ ] CI green on \`x86_64-pc-windows-msvc\` (static link path on MSVC)
- [ ] Reviewer smoke-test on a system without CUDA → graceful fallback

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,11 +102,29 @@ jobs:
       # adds cl.exe to PATH (required by nvcc on Windows to compile CUDA kernels).
       - name: Install CUDA toolkit (Windows x86_64)
         if: matrix.target == 'x86_64-pc-windows-msvc'
+        id: cuda-toolkit-windows
         uses: Jimver/cuda-toolkit@v0.2.22
         with:
           cuda: '12.6.0'
           method: network
           sub-packages: '["nvcc", "cudart", "nvrtc", "nvrtc_dev", "cublas", "cublas_dev", "curand", "curand_dev"]'
+
+      # Jimver/cuda-toolkit already sets CUDA_PATH in the runner environment,
+      # but pin it explicitly so candle-kernels/build.rs can't silently fall
+      # back to its hardcoded default if a future version of the action
+      # changes that behavior.  Fail fast with a clear message if CUDA_PATH
+      # is somehow unset — otherwise the link step will later fail deep in
+      # nvcc output with undefined cudart_static symbols.
+      - name: Export CUDA_PATH for build.rs (Windows x86_64)
+        if: matrix.target == 'x86_64-pc-windows-msvc'
+        shell: pwsh
+        run: |
+          if (-not $env:CUDA_PATH) {
+            Write-Error "CUDA_PATH is not set after Jimver/cuda-toolkit install"
+            exit 1
+          }
+          Write-Host "CUDA_PATH=$env:CUDA_PATH"
+          echo "CUDA_PATH=$env:CUDA_PATH" >> $env:GITHUB_ENV
 
       - name: Set up MSVC environment (Windows x86_64 — nvcc needs cl.exe)
         if: matrix.target == 'x86_64-pc-windows-msvc'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,9 +96,6 @@ jobs:
           sudo ln -sfn /usr/local/cuda/lib64/stubs/libcuda.so \
                        /usr/local/cuda/lib64/stubs/libcuda.so.1
           echo "LD_LIBRARY_PATH=/usr/local/cuda/lib64/stubs:${LD_LIBRARY_PATH}" >> "$GITHUB_ENV"
-          # candle-kernels/build.rs emits `cargo:rustc-link-lib=dylib=cudart`.
-          # The linker needs to find libcudart.so which lives in lib64 (not stubs).
-          echo "LIBRARY_PATH=/usr/local/cuda/lib64:${LIBRARY_PATH}" >> "$GITHUB_ENV"
 
       # Install CUDA toolkit and set up MSVC environment on Windows x86_64.
       # Jimver/cuda-toolkit installs nvcc and adds it to PATH; ilammy/msvc-dev-cmd
@@ -114,16 +111,6 @@ jobs:
       - name: Set up MSVC environment (Windows x86_64 — nvcc needs cl.exe)
         if: matrix.target == 'x86_64-pc-windows-msvc'
         uses: ilammy/msvc-dev-cmd@v1
-
-      # candle-kernels/build.rs emits `cargo:rustc-link-lib=dylib=cudart`.
-      # The MSVC linker resolves .lib files via the LIB env var; expose the
-      # CUDA lib/x64 directory so cudart.lib can be found at link time.
-      - name: Add CUDA lib to MSVC linker search path (Windows x86_64)
-        if: matrix.target == 'x86_64-pc-windows-msvc'
-        shell: pwsh
-        run: |
-          $cudaLib = "$env:CUDA_PATH\lib\x64"
-          echo "LIB=$cudaLib;$env:LIB" >> $env:GITHUB_ENV
 
       # Git ships its own link.exe in usr/bin which shadows MSVC's link.exe
       # when bash shells are used. Rename it so the MSVC linker is found first.
--- a/candle-kernels/build.rs
+++ b/candle-kernels/build.rs
@@ -56,6 +56,7 @@ fn main() {
     // This is what makes "brew install inferrs" viable as a single binary
     // that dlopens whatever CUDA libs are present at runtime (12.x, 13.x, …)
     // and falls back cleanly on systems without CUDA at all.
+    //
     // `rustc-link-lib=static=` propagates from lib build scripts to downstream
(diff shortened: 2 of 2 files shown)


igorls:
Thanks for the review. A note on scope for the human reviewers:

The `float8/` directory in this PR is an **unmodified copy of upstream [float8 v0.6.1](https://github.com/EricLBuehler/float8)**, with a single-line change in `float8/Cargo.toml` (`cudarc?/dynamic-linking` → `cudarc?/dynamic-loading`). That one-line delta is the entire point of the PR — it fixes the CUDA startup crash described in the PR body.

All issues flagged in the bot review — NaN handling in `min` / `max` / `classify`, doc typos, the assertion-less roundtrip test, the inconsistent `DeviceRepr` path — are **pre-existing bugs in upstream float8 v0.6.1**, not introduced by this PR.

We're keeping the vendored source pristine intentionally:

1. **Minimal diff from upstream** — this makes the vendored copy easy to drop once cudarc's `dynamic-loading` support propagates through a released float8, so we don't carry this patch forever.
2. **Upstream is the right place** for general correctness fixes. They affect every float8 consumer, not just inferrs. Patching them here only helps us and creates drift we'd have to re-reconcile on every float8 bump.
3. **Small blast radius** — a one-line feature-flag change is trivial to review and revert. Bundling unrelated correctness fixes would expand the review surface for no inferrs-side benefit (we don't hit the flagged NaN min/max paths in any hot code).

If any of these turn out to affect inferrs in practice, we'll file them upstream at https://github.com/EricLBuehler/float8 rather than carry a growing patch set locally.

ericcurtin:
@igorls thanks for the PR. I agree you have made the right choices here. You have conflicts to fix and I think this should be good to merge

igorls:
I see float8 was updated to 0.7.0, I'm revalidating this on my system

ericcurtin:
And thanks for looking at this, I tried to fix for a little bit but I was doing like 5 things at once. Just so you know the idea is we can:

brew install inferrs

and it will dynamically load any cuda lib available 12.9, 13.0, 13.1, etc. The goal is one binary for all. And after that the full fallback sequence is:

CUDA, ROCm, Metal, Hexagon, OpenVino, MUSA, CANN, Vulkan, CPU

ericcurtin:
main branch is failing on CUDA also you may want to look at that, it's related

igorls:
> main branch is failing on CUDA also you may want to look at that, it's related

yes this was the main issue I was having on CUDA 13.1

with my patch I've started to test using gemma-4-E2B-it against native ollama and spotted many problems regarding streaming and reasoning decoding (was working on fixing them)

ericcurtin:
> > main branch is failing on CUDA also you may want to look at that, it's related
> 
> yes this was the main issue I was having on CUDA 13.1
> 
> with my patch I've started to test using gemma-4-E2B-it against native ollama and spotted many problems regarding streaming and reasoning decoding (was working on fixing them)

Yeah feel free to open PRs for any of those things in here.

You could try and open PRs in Ollama but they are not spectacular at looking at random PRs, it's mainly Ollama employee's PRs they are interested in.

Also Ollama just copy what llama.cpp do generally so normally if you fix it in llama.cpp it propagates to Ollama as a better approach to enhancing Ollama.

But this is it's own Rust implementation and IMO a cleaner, better long-term design

ericcurtin:
@igorls build failed, please look at the bot comments also

igorls:
Fixed the aarch64 link failure plus both Copilot review comments in 8aac711 (on top of 2d4ccdf which fixed the underlying aarch64 issue).

**Root cause of the red build** (aarch64-unknown-linux-gnu test link): `rustc-link-search=native=$CUDA_PATH/lib64` is Debian/x86_64-specific. On the sbsa cuda-toolkit-12-6 package (used by `ubuntu-24.04-arm` runners), `libcudart_static.a` lives under `targets/sbsa-linux/lib/` and there's no `lib64` → `targets/…` compatibility symlink. Rustc silently couldn't find the archive, dropped `-l cudart_static` from the linker command, and the link failed with a flood of undefined references out of `libmoe.a` (`cudaLaunchKernel_ptsz`, `__cudaRegisterFatBinary`, etc.). 2d4ccdf probes all plausible layouts (`lib64`, `lib`, `targets/x86_64-linux/lib`, `targets/sbsa-linux/lib`, `targets/aarch64-linux/lib`) and registers every existing one as a native search path, so rustc resolves the archive regardless of which layout the host toolkit ships.

**Bot comment #1** (candle-kernels/build.rs — `rustc-link-lib=static=cudart_static` always emitted but `rustc-link-search=` conditional on `CUDA_PATH`): 8aac711 drops the "only if CUDA_PATH is set" gate and probes sensible defaults when the env var is missing — `/usr/local/cuda` on Unix; a newest-first list of `NVIDIA GPU Computing Toolkit\CUDA\v13.2..v12.0` on Windows. The warning emitted when nothing is found now includes a concrete hint to set `CUDA_PATH` with a per-platform example.

**Bot comment #2** (ci.yml Windows CUDA_PATH implicit dependency on the Jimver action): 8aac711 adds an explicit "Export CUDA_PATH for build.rs" step right after the Jimver install that (a) errors out loudly if `CUDA_PATH` is unset, and (b) re-emits it into `$GITHUB_ENV` so later steps see it unconditionally. No-op on the happy path; fails fast with a clear message on the unhappy one, instead of surfacing as a deep nvcc/linker error.

I reproduced the aarch64 CI failure locally with `cargo test --no-run -p inferrs` (debug build, where `--gc-sections` can't strip libmoe.a because the test targets keep its symbols alive). With both fixes applied, the unit-test binary and `server_integration` test binary link cleanly and `readelf -d` shows zero CUDA libs in DT_NEEDED — same as the release binary.

The earlier bot comments on the pre-pivot PR version (the ones about `float8/src/lib.rs` NaN handling, `classify` ordering, and doc typos) aren't relevant anymore — I dropped the vendored `float8/` directory entirely in the pivot since your 646375b upgrade to float8 0.7.0 superseded that approach.

PR #156, opened 2026-04-10
feat: Ollama API parity — thinking mode + duration metrics

igorls:
## Summary

Full Ollama API parity for thinking/reasoning models and duration metrics.

### What this PR does

**1. Thinking mode (`think: true/false`)**
- Auto-detects thinking-capable models (e.g. Gemma4) via `<|think|>` token presence in the vocabulary
- `/api/chat` defaults to `think=true` when the model supports it, matching Ollama behavior
- `/api/generate` keeps `think=false` default to avoid breaking raw prompts
- When `think=true`: reasoning surfaces in a separate `thinking` field
- When `think=false`: reasoning tokens fold back into `content` (nothing silently dropped)
- Engine-level `ThinkFilter` classifies tokens at the token ID level — both streaming and blocking paths share the same code

**2. Duration metrics**
- `total_duration`, `load_duration`, `prompt_eval_duration`, `eval_duration` on all `/api/generate` and `/api/chat` responses (streaming + non-streaming)
- `prompt_eval_count` and `eval_count` populated correctly on streaming final chunks
- Timings measured at the engine level using wall-clock `Instant` to avoid HTTP/queue transport delay

**3. OllamaOptions extensions**
- Accepts `seed`, `min_p`, `stop`, `presence_penalty`, `frequency_penalty`
- `stop` tokens are resolved via tokenizer and wired to `extra_stop_token_ids`

**4. Test suite**
- `tests/compare_ollama.sh`: 9-test parity suite covering:
  - Non-streaming chat with default/think=true/think=false
  - Streaming chat with thinking chunk validation
  - Non-streaming and streaming `/api/generate`
  - Schema metric checks (done_reason, eval_count, prompt_eval_count, total_duration)
  - Degeneration detection
  - Thinking field quality (minimum length check)

### Tested with
- `gemma4:e2b` (Ollama) vs `google/gemma-4-E2B-it` Q4K (inferrs)
- **9/9 tests pass** with full schema and behavioral parity

### Commits
1. `feat: ollama duration timings + thinking/reasoning mode` — core implementation
2. `fix: think mode activation and ThinkFilter token detection` — Gemma4 tokenizer fallbacks
3. `feat: auto-detect thinking models + streaming metrics parity` — default think=true for capable models

### DIFF
--- a/inferrs/src/engine.rs
+++ b/inferrs/src/engine.rs
@@ -16,6 +16,7 @@
 
 use std::collections::VecDeque;
 use std::sync::{Arc, Mutex};
+use std::time::Instant;
 
 use anyhow::Result;
 use candle_core::{DType, Device, Tensor};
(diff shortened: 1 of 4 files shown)


igorls:
Happy to rebase and adapt once https://github.com/ericcurtin/inferrs/pull/155 is merged

ericcurtin:
@igorls feel free to rebase and adapt, also address the bot comments

PR #158, opened 2026-04-11
Split VisionConfig into enum with per-family variants to handle Qwen 3.5

co-esquie:
After reviewing https://github.com/ericcurtin/inferrs/pull/157, Fixes a crash when loading Qwen3.5 models by refactoring `VisionConfig` from a single flat struct into an enum with per-family variants (`Gemma4` / `Qwen`), enabling proper support for Qwen3.5's different vision encoder config shape.

### Problem

On `main`, loading a Qwen3.5 model would crash because `vision_config` in Qwen3.5's `config.json` has a completely different schema than Gemma4's (e.g. `depth` instead of `num_hidden_layers`, no `pooling_kernel_size`, etc.). The single `VisionConfig` struct couldn't deserialize it.

**Broken on `main`:**
```bash
./target/release/inferrs run Qwen/Qwen3.5-0.8B "Describe docker company in one sentence" --device=cuda
```

> Docker is a software development platform that enables developers to build, run, and deploy applications across multiple environments using containerized code.

### DIFF
--- a/inferrs/src/config.rs
+++ b/inferrs/src/config.rs
@@ -66,9 +66,49 @@ pub struct TextConfig {
 }
 
 /// Vision encoder configuration from `vision_config` in `config.json`.
+///
+/// Different model families ship different vision encoders, each with its own
+/// config shape.  The `model_type` field inside `vision_config` is used to
+/// dispatch to the right variant at deserialization time.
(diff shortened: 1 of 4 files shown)


co-esquie:
Greptile feedback applied

PR #162, opened 2026-04-11
Use candle-nn kernels + enable TurboQuant for Qwen 3.5

co-esquie:
Replace hand-rolled RMS normalization (rms_norm_tensor) and sigmoid in Qwen3.5 linear attention with candle_nn::ops built-in kernels. Removes 20 LOC and leverages optimized Metal/CUDA kernels already available in candle-nn.

Enable TurboQuant KV cache compression for Qwen3.5 — the implementation was already wired in the model, only the architecture guard in models/mod.rs was missing from the match arm.

### Current State of Qwen 3.5 0.8B (inferrs-benchmark)

| Métrique      | inferrs     | llama.cpp   |
|---------------|-------------|-------------|
| TTFT          | 939 ms      | 51 ms       |
| Prefill       | 320 tok/s   | 5783 tok/s  |
| Decode        | 17 tok/s    | 48 tok/s    |
| Peak memory   | 630 MB      | 4079 MB     |

Current implementation is really slow since it's a naive implementation and must be considered as WIP state. There is few issues I'm working on to improve performances compared to llama cpp

### DIFF
--- a/inferrs/src/models/mod.rs
+++ b/inferrs/src/models/mod.rs
@@ -519,11 +519,11 @@ pub fn load_model(
     // TurboQuant is on by default; warn if this architecture doesn't support it.
     if turbo_quant_bits.is_some() {
         match arch {
-            ModelArchitecture::Qwen3 | ModelArchitecture::Gemma4 => {} // supported
+            ModelArchitecture::Qwen3 | ModelArchitecture::Qwen35 | ModelArchitecture::Gemma4 => {}
             other => {
                 tracing::warn!(
                     "--turbo-quant is not supported for {:?} and will be ignored. \
-                     TurboQuant KV cache compression is currently only available for Qwen3 and Gemma4. \
+                     TurboQuant KV cache compression is currently only available for Qwen3, Qwen3.5, and Gemma4. \
                      Pass --turbo-quant=false to suppress this warning.",
                     other
                 );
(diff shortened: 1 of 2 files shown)


co-esquie:
Greptile is right as always

ericcurtin:
Btw, turboquant should be on by default everywhere, although it's possible to turn it off

ericcurtin (approved):

ericcurtin:
I'm probably gonna turn on quantize by default also. Because it will leave a bad first impression if inferrs isn't superfast by default like llama.cpp . People interested in production like workloads can use something like --paged-attention

ericcurtin (approved):

ericcurtin:
@co-esquie you won't beat llama.cpp without using --quantize by the way

you should also measure --paged-attention vs vLLM, optimize for both

co-esquie:
> @co-esquie you won't beat llama.cpp without using --quantize by the way
> 
> you should also measure --paged-attention vs vLLM, optimize for both

I wasn't expecting to beat llama cpp but current implementation is very naïve and lacks plenty of optimisations for this architecture. I'll run better benchmarks when implementation will be more mature, just wanted to point for the users it's far behind Gemma 4 progress

ericcurtin:
> > @co-esquie you won't beat llama.cpp without using --quantize by the way
> > you should also measure --paged-attention vs vLLM, optimize for both
> 
> I wasn't expecting to beat llama cpp but current implementation is very naïve and lacks plenty of optimisations for this architecture. I'll run better benchmarks when implementation will be more mature, just wanted to point for the users it's far behind Gemma 4 progress

You will beat it if you keep going with --quantize , it's not as hard as you think. I beat llama.cpp with Gemma 4 and --quantize .

co-esquie:
I think even quantified version will be slower ☠️

ericcurtin:
> I think even quantified version will be slower ☠️

Well if you give up yes ;)

PR #163, opened 2026-04-11
perf: pre-compute conv1d_weight (F32) and lm_head transposition at model load

co-esquie:
## Perf: pre-compute conv1d F32 cast & lm_head transpose at model load

Two redundant operations moved from the forward path to model construction:
- BF16→F32 dtype cast of conv1d weight in linear attention layers (18 layers × per forward on Qwen3.5)
- Transpose + contiguous materialization of lm_head (~300MB per forward on Qwen3.5-0.8B)

Affects Qwen3 and Qwen3.5. Gemma4 unaffected (QLinear for lm_head, no conv1d).

**Expected gains:** Eliminates per-layer GPU dispatch overhead on every prefill + a full vocab-sized allocation on every decode step.


co-esquie:
feedbacks applied

PR #166, opened 2026-04-11
refactor(qmatmul): extract QLinear/QGgufVarBuilder and wire Qwen3.5

co-esquie:
# QMatMul for Qwen3.5

## Problem

Qwen3.5 does not benefit from the `--quantize` flag at decode time. When quantization is requested, inferrs converts model weights to GGUF format — but immediately dequantizes every tensor back to BF16 on load. All linear projections run as standard BF16 GEMV during decode, so quantization only saves disk space, not GPU memory bandwidth or compute time.

Decode is the phase where quantized weights matter most: it is memory-bound (single-token GEMV), so keeping weights compressed in GPU memory directly reduces bandwidth pressure and yields a 2-3× throughput gain. Prefill is compute-bound and benefits much less.

Gemma4 already has this working: its projection layers use a `QMatMul`-backed path that dispatches to Metal/CUDA quantized GEMV kernels on compressed weights directly — the same strategy as llama.cpp.

## Preliminary Work

This PR lays the groundwork for bringing the same quantized decode path to Qwen3.5.

**Shared quantization module.** `QLinear`, `QGgufVarBuilder` and the `qlinear_b` helper were extracted from `gemma4.rs` into a new shared `quantized_linear` module, making them available to any model architecture.

**Wiring.** `QGgufVarBuilder` is now constructed for Qwen3.5 in `mod.rs` alongside Gemma4, and threaded down through the full model hierarchy (`Qwen35Model` → `DecoderLayer` → `FullAttention` / `LinearAttn`).

The plumbing is in place but the quantized path is not yet active: projections still call the standard `linear_no_bias` loader. The next step is to swap those calls for `qlinear_b`, which will keep weights in their compressed form and route GEMV through the quantized Metal/CUDA kernel during decode.


co-esquie:
All bots feedbacks applied, would appreciate if it's possible to relaunch them the regression potential is real

PR #167, opened 2026-04-11
Fix: GGUF Tokenizer Source & LLaMa.cpp Tensor Mappings

igorls:
## Description
This PR implements two major fixes for GGUF model handling in `inferrs`:
1. **Added `--tokenizer-source` CLI flag**: Allows manually overriding the tokenizer and HF config source. GGUF-only repos (like Unsloth) usually lack the required `tokenizer.json` files or GGUF metadata (`general.source.repo_id`). This explicitly instructs the daemon to pull the tokenizer config from the canonical upstream model.
2. **Applied Safetensors-to-GGUF tensor mapping logic for Phi3**: `inferrs` expects Safetensors standard variable names because internal GGUF conversions retain them. However, Llama.cpp standardizes them to a different spec (e.g. `model.embed_tokens.weight` mapping to `token_embd.weight`). When an external GGUF is detected, `inferrs` now automatically maps these Llama.cpp queries back to HF safetensors equivalents on-the-fly for `Phi3`.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Testing
- Verified successful `serve` inference execution against `unsloth/Phi-4-reasoning-plus-GGUF` combined with `--tokenizer-source microsoft/Phi-4-reasoning-plus`. Models initialize properly and `llama.cpp` tensor mappings are correctly applied.

### DIFF
--- a/inferrs/src/engine.rs
+++ b/inferrs/src/engine.rs
@@ -134,6 +134,7 @@ pub fn load_engine(args: &ServeArgs) -> Result<EngineContext> {
         model_id,
         &args.revision,
         args.gguf_file.as_deref(),
+        args.tokenizer_source.as_deref(),
         quant_dtype,
     )?;
 
--- a/inferrs/src/hub.rs
+++ b/inferrs/src/hub.rs
@@ -86,6 +86,7 @@ pub fn download_model(
     model_id: &str,
     revision: &str,
     gguf_file: Option<&str>,
+    tokenizer_source: Option<&str>,
 ) -> Result<ModelFiles> {
     // If the model_id looks like a local path, load directly without network.
     let as_path = std::path::Path::new(model_id);
@@ -108,7 +109,7 @@ pub fn download_model(
 
     // Fast-path: caller explicitly asked for a specific GGUF file.
     if let Some(fname) = gguf_file {
-        return download_gguf_only_repo(&repo, &api, model_id, fname);
+        return download_gguf_only_repo(&repo, &api, model_id, fname, tokenizer_source);
     }
 
     // Probe for config.json.  If it is missing the repo is likely GGUF-only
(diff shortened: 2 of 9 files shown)


ericcurtin:
The bot comments are worth addressing

jexp:
Seems we worked on the same thing, I was coming from existing Ollama downloaded GGUF files :)

https://github.com/jexp/inferrs/tree/feature/ollama-integration-rebased

Compared with igor/fix/gguf-tokenizer-source:

- No `--tokenizer-source` flag; igor adds an explicit HF repo override
  for GGUF-only repos that lack `general.source.repo_id` metadata
- Per-arch rename functions in `gguf.rs` with unit tests; igor has a
  unified `gguf_rename_tensor(name, arch)` dispatch table in `mod.rs`
- Lazy `rename` field on `QGgufVarBuilder` applied per `get()`; igor
  eagerly re-keys all tensors via `rename_keys()` at construction time
- Explicit `GgufNaming::LlamaCpp` guard to skip vision encoder load;
  igor catches `"cannot find tensor"` error strings at load time
- `contains_key` guard for PLI dual naming in `gemma4.rs`; igor uses
  an `or_else` fallback — functionally equivalent
- No `tracing::debug!` for top GGUF keys; igor logs the first 10

igorls:
All bot review comments have been addressed in commit 889eb9a:

1. **Greptile: asymmetric `gate_up_proj`→`ffn_up` reverse mapping** — Fixed. Reverse mapping is now arch-aware: Phi3 → `mlp.gate_up_proj.weight`, others → `mlp.up_proj.weight`.

2. **Copilot: GGUF key dump at INFO** — Downgraded to `tracing::debug!`.

3. **Copilot: `ffn_up` reverse mapping** — Same fix as #1.

4. **Copilot: audio/vision encoder error masking** — Now only skips on `"cannot find tensor"` errors; other failures (corrupted tensors, shape mismatches) propagate normally.

5. **Copilot: `rename_keys` eager load + error suppression** — Errors now propagate (`-> Result<Self>`), locks are no longer held simultaneously.

6. **Copilot: `tokenizer_source` scope** — By design: only relevant for GGUF-only repos that lack tokenizer files. Non-GGUF repos already have their own tokenizer.

7. **Gemini: unrecognized tensor fallback** — Already handled: unknown suffixes pass through with `blk.{idx}.` prefix applied.

ericcurtin:
@jexp I'm not sure about adding support for Ollama models. Why not just use Ollama in that case?

PR #169, opened 2026-04-11
convert Qwen3.5 to QLinear for all projections and lm_head

co-esquie:
### Functional

Qwen3.5 was loading all GGUF weights as dequantized BF16, losing the benefit of quantized GEMV kernels during decode. This PR converts all 12 linear projections (attention + MLP across 24 layers) and the lm_head to use the `QLinear`/`QMatMul` path already shared with Gemma4. Decode throughput (memory-bound GEMV) should improve ~2-3x. Prefill (compute-bound GEMM) sees marginal benefit. Weight tying with embed_tokens is preserved — lm_head falls back to online Q4K quantization for safetensors models.

### Technical

- **Projections converted**: FullAttention (q/k/v/o_proj), LinearAttn (in_proj_qkv/z/a/b, out_proj), and a local `QMlp` struct (gate/up/down_proj) — all `linear_no_bias()` → `qlinear_b()`.
- **lm_head**: `Tensor` → `QLinear` following Gemma4's pattern — GGUF quantized weight first, online Q4K for safetensors, BF16 fallback.
- **Gate**: `QGgufVarBuilder` enabled for `ModelArchitecture::Qwen35` in `mod.rs`, so the quantized GGUF path is now active.
- `QMlp` is defined locally in `qwen3_5.rs` rather than modifying the shared `Mlp` in `attention_utils.rs` (still used by Qwen3 with plain `Linear`).

### DIFF
--- a/inferrs/src/models/mod.rs
+++ b/inferrs/src/models/mod.rs
@@ -506,7 +506,10 @@ pub fn load_model(
     // NOTE: Only enabled for architectures that have been updated to use
     // qlinear_b. Enabling it before the model uses QLinear would load weights
     // twice (quantized + dequantized), doubling memory usage.
-    let qvb: Option<QGgufVarBuilder> = if matches!(arch, ModelArchitecture::Gemma4) {
+    let qvb: Option<QGgufVarBuilder> = if matches!(
+        arch,
+        ModelArchitecture::Gemma4 | ModelArchitecture::Qwen35
+    ) {
         gguf_path.and_then(|p| {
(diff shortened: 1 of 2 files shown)


co-esquie:
Bots are fully rights and feedbacks are applied. There is still some work to bring qwen 3.5 on par with gemma 4 but the numbers when running benchmark (which use --quantitize by default) look promising.

ericcurtin:
Yeah you can't really beat --quantize (it's llama.cpp style essentially) so it's likely we turn it on by default, it's just in a server environment you probably don't want that you want --paged-attention

Issue #173, opened 2026-04-12
Gemma 4 26B MoE (A4B) produces gibberish output — expert routing issue in GGUF


iguinea:
## Description

The Gemma 4 26B MoE model (`google/gemma-4-26B-A4B-it`) produces completely incoherent output (random mixed-language characters) when served via inferrs with GGUF quantization. The dense Gemma 4 31B model works perfectly with the same setup, suggesting an issue with the MoE expert routing implementation in the GGUF path.

## Environment

- **OS:** macOS 15 (Darwin 25.3.0)
- **Hardware:** MacBook Pro, Apple M4 Max, 128 GB unified memory
- **inferrs version:** installed via Homebrew (`/opt/homebrew/bin/inferrs`)
- **Device:** Metal
- **Available RAM at time of test:** 117+ GB (confirmed via `llmfit system`)

### 1. Pull and quantize the 26B MoE model

```bash
inferrs pull google/gemma-4-26B-A4B-it --quantize=Q8_0
```

Output: `Saved 25.49 GiB → .../model-Q8_0.gguf`

### 2. Serve the model

```bash
inferrs serve google/gemma-4-26B-A4B-it --device metal --quantize=Q8_0 --turbo-quant=8 --port 17434
```

Model loads successfully:
```
INFO inferrs::models  Model loaded successfully
INFO inferrs::models  Gemma4 config: 30 layers, 16 heads, 2816 hidden, 8 kv_heads
INFO inferrs::models::gemma4  TurboQuant KV cache enabled: 8 bits/coord
INFO inferrs::server  Server listening on 0.0.0.0:17434
INFO inferrs::engine  Engine loop started (continuous batching, max_batch_size=1, paged=false)
```

### 3. Send a simple request

```bash
curl -s http://127.0.0.1:17434/v1/chat/completions -X POST \
  -H "Content-Type: application/json" \
  -d '{
    "messages": [
      {"role": "system", "content": "Extract concepts as JSON: [{\"name\":\"x\",\"definition\":\"y\"}]"},
      {"role": "user", "content": "Claude Code is a CLI by Anthropic for terminal-based LLM work. It supports Skills and MCP."}
    ],
    "max_tokens": 200,
    "stream": false,
    "temperature": 0.2
  }'
```

### 4. Observe garbage output

Response (prompt=70 tokens, completion=500 tokens, **6.5 seconds** — fast):

```
[...]

PR #175, opened 2026-04-12
fix: add GemmaNormFixBackend to correct RMSNorm weights for Gemma2/3 external GGUFs

jexp:
llama.cpp stores the actual RMSNorm scale value directly in GGUF files, but candle-transformers expects scale - 1.0 (HuggingFace convention). 

This adds GemmaNormFixBackend wrapping GgufBackend that subtracts 1.0 from any tensor whose name ends in "norm.weight", applied only for Gemma2/Gemma3 external GGUFs.

Could perhaps also be the reason for https://github.com/ericcurtin/inferrs/issues/173

### DIFF
--- a/inferrs/src/models/mod.rs
+++ b/inferrs/src/models/mod.rs
@@ -436,15 +436,64 @@ impl candle_nn::var_builder::SimpleBackend for GgufBackend {
     }
 }
 
+/// A [`candle_nn::var_builder::SimpleBackend`] that wraps [`GgufBackend`] and
+/// subtracts 1.0 from any tensor whose name ends in `"norm.weight"`.
+///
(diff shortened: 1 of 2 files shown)


ericcurtin:
Lets test this and not assume:

Could perhaps also be the reason for https://github.com/ericcurtin/inferrs/issues/173

Btw, that would be a better one to look at than any other additional tests. This is something we know is not working today.

jexp:
@ericcurtin tested it locally with a Gemma4 E2B GGUF (non-ollama, and there it was fine).

I'm back home now, so I can test the Gemma4 21b MoE model now as well.

jexp:
Unrelated to #173 - that had just a the missing MoE implementation - current implementation only had the dense layer part up until the router. See #181 

So the RMSNorm fix only applies to Gemma2/3

PR #177, opened 2026-04-12
fix: reset stale Loading slot on sender drop; accept any model name for single-model server

jexp:
Two fixes to src/server.rs that came up during integration testing. Can split it into two PRs if required.

1. `wait_for_slot_load()` now takes &AppState and resets `ModelSlot::Loading` → Empty when the watch Sender is dropped (background thread panicked without publishing a result). This prevents /health returning 503 forever. Returns 503 `SERVICE_UNAVAILABLE` with a "please retry" message instead of 500.

2. `resolve_openai_model()` gains a single-model fast path: when the server was started with an explicit --model argument, the client's requested model name is accepted unconditionally and the preloaded model is returned. Generic clients that always send e.g. "model": "gpt-4" now work without error.

### DIFF
--- a/inferrs/src/server.rs
+++ b/inferrs/src/server.rs
@@ -1294,9 +1294,10 @@ async fn load_model_on_demand(
                 rx,
             } if model_matches_id(loading_id, model_id) => {
                 // Another task is already loading this model — wait on its result.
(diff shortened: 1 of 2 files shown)


jexp:
I'll address the feedback.

PR #188, opened 2026-04-13
Qwen3.5 startup latency (online quantization bottleneck)

co-esquie:
## Problem

Serving a Qwen3.5 model from safetensors triggers an online BF16→Q4K quantization at startup. On the 0.8B model this takes ~19s; on the 2B model ~52s on my hardware. During that time the GPU sits idle and the server is unresponsive.

The bottleneck is 187 sequential CPU quantizations (one per linear projection + lm_head), each doing a GPU→CPU copy, a tight loop over 256-element blocks, then a CPU→GPU copy back.

## Fix

Two targeted changes, no API surface affected:

**Parallelize layer construction** (`qwen3_5.rs`) — the 24 decoder layers are independent of each other, and lm_head depends only on the embedding which is already loaded. A `rayon::join` runs layer construction (`par_iter` over all layers) and lm_head quantization concurrently. `VarBuilder` is `Send`, so no restructuring of constructors was needed.

**Parallelize block quantization** (`candle-core/k_quants.rs`) — `BlockQ4K::from_float` iterated sequentially over 256-element blocks. `group_for_quantization` already returns a `Vec` of disjoint `(&mut block, &[f32])` pairs; replacing the `for` loop with `into_par_iter().for_each` is the entire change. This directly benefits lm_head, which quantizes a single large tensor (254M–509M elements) and can now use the threads freed by the layer pass.

## Result on my hardware

| Model | Before | After |
|---|---|---|
| 0.8B | ~19s | ~3-4s |
| 2B | ~52s | ~6-8s |

### DIFF
--- a/candle-core/src/quantized/k_quants.rs
+++ b/candle-core/src/quantized/k_quants.rs
@@ -1452,63 +1452,65 @@ impl GgmlType for BlockQ4K {
     }
 
     fn from_float(xs: &[f32], ys: &mut [Self]) {
-        for (block, x) in group_for_quantization(xs, ys) {
-            let mut mins: [f32; QK_K / 32] = [0.0; QK_K / 32];
-            let mut scales: [f32; QK_K / 32] = [0.0; QK_K / 32];
+        group_for_quantization(xs, ys)
+            .into_par_iter()
+            .for_each(|(block, x)| {
+                let mut mins: [f32; QK_K / 32] = [0.0; QK_K / 32];
+                let mut scales: [f32; QK_K / 32] = [0.0; QK_K / 32];
 
-            for (j, x_scale_slice) in x.chunks_exact(32).enumerate() {
-                (scales[j], mins[j]) = make_qkx1_quants(15, 5, x_scale_slice);
-            }
-
-            // get max scale and max min and ensure they are >= 0.0
-            let max_scale = scales.iter().fold(0.0, |max, &val| val.max(max));
-            let max_min = mins.iter().fold(0.0, |max, &val| val.max(max));
-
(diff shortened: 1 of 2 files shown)


co-esquie:
Marketing can now claim inferrs is twice as fast as LM studio (which use llama cpp) for qwen models (in decode only). There is more to come but it has been a long journey to have this results

```
cargo run --release --bin inferrs-benchmark -- --inferrs-model Qwen/Qwen3.5-2B --llama-model lmstudio-community/Qwen3.5-2B-GGUF

═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
Backend                                                           TTH (ms)     TTFT (ms)   Prefill (t/s)   Decode (t/s)   Peak mem (MB)
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
llama-server -hf lmstudio-community/Qwen3.5-2B-GGUF            16145.82 ms      82.98 ms     3599.72 t/s      25.11 t/s      5820.57 MB
inferrs serve --quantize Qwen/Qwen3.5-2B                        6029.67 ms     197.48 ms     1524.45 t/s      57.15 t/s       927.30 MB
inferrs serve --turbo-quant=false --quantize Qwen/Qwen3.5-2B    4004.46 ms     197.61 ms     1523.38 t/s      58.83 t/s       923.94 MB
inferrs serve Qwen/Qwen3.5-2B                                  16038.16 ms     227.97 ms     1324.73 t/s      53.45 t/s      1012.02 MB
═══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════
```

ericcurtin (approved):

PR #211, opened 2026-04-15
Feat/repetition penalty flags Fixes #183

### LINKED ISSUE #183: Repetition/degeneration in long generations
## Problem

Long generations (roughly >512 tokens) from sliding-window models like Gemma4 26B A4B can degrade into repetitive loops — repeated words joined by hyphens, or a single token repeated indefinitely. This was observed with `inferrs run google/gemma-4-26B-A4B-it --quantize=Q8_0`.

The root cause is a combination of:

1. **Sliding-window context loss** — Gemma4 26B has a 512-token sliding window on most attention layers. After 512 decode steps the model loses access to the structural context from the beginning of its own output, making local repetition attractors more likely.
2. **No repetition penalty applied by default** — inferrs defaults to `repetition_penalty = 1.0` (disabled) and `frequency_penalty = 0.0` (disabled), so nothing breaks the feedback loop once it starts.

---

## Current state

The sampler (`sampler.rs`) already implements:
- `repetition_penalty` — llama.cpp-style: divides positive logits / multiplies negative logits of previously-seen tokens (deduped via HashSet, so each unique token is penalised once)
- `frequency_penalty` — OpenAI-style: subtracts `penalty × count` per token
- `presence_penalty` — OpenAI-style flat per-token penalty

These are exposed on the OpenAI `/v1/chat/completions` and Ollama `/api/chat` server endpoints. The `run` CLI previously had no flags for them (fixed in #TODO — adds `--repetition-penalty` and `--frequency-penalty` to `RunArgs`).

---

## What is missing: `repeat_last_n`

The sampler currently applies the penalty against the **entire token history** (`previous_tokens`). llama.cpp and Ollama limit the look-back to a configurable window via `repeat_last_n` (default 64 in both).

This matters because:

- Without a window, `repetition_penalty = 1.2` divides the logit of *every* token the model has ever produced — including words from the opening paragraph of a 3,000-token essay. Legitimate vocabulary gets suppressed, leading to awkward synonym-reaching.
- With `frequency_penalty`, the situation is worse: the penalty is `fp × count`. At 2,000 tokens, "the" appears ~150 times → `0.15 × 150 = 22.5` logit subtracted. That effectively bans the word for the rest of the generation.

`repeat_last_n` confines the penalty to recent tokens (e.g. the last 64), which specifically targets the local looping pattern without harming vocabulary diversity at long range.

---

## What other tools do

| Tool | Default repetition penalty | Window |
|------|---------------------------|--------|
| Ollama / llama.cpp | `repeat_penalty = 1.1` | `repeat_last_n = 64` |
| vLLM | `repetition_penalty = 1.0` (off) | full history |
| OpenAI API | `frequency_penalty = 0.0` (off) | full history |
| text-generation-webui | ~1.18 (community) | full history |

Ollama's approach is the most practical: a mild `1.1` penalty over a short recent window is enough to break local loops without degrading long-range coherence.

---

## Default mismatch with Ollama

inferrs aims for Ollama API compatibility, but its default `repetition_penalty = 1.0` diverges from Ollama's `1.1`. Clients that omit the parameter get different behaviour from the two servers.

---

### 1. Add `repeat_last_n` to `SamplingParams` and the sampler
Limit the penalty look-back to the last N tokens. Wire through `OllamaOptions` (`repeat_last_n`) and the OpenAI/Ollama endpoints.

### 2. Change `SamplingParams::default()` to `repetition_penalty = 1.1`
Match Ollama's default. Safe to do even without `repeat_last_n` because the HashSet deduplication means each unique token is penalised at most once — a 1.1 divisor on 2,000-token history is tolerable, unlike frequency_penalty which compounds.

### 3. Keep `frequency_penalty = 0.0` as the default
The count-based compounding makes any non-zero value harmful for outputs longer than a few hundred tokens without a `repeat_last_n` window. Do not set a non-zero default here until `repeat_last_n` is implemented.

### 4. Document the sliding-window degeneration pattern
Add a note to the README / model docs that sliding-window models (Gemma4 26B, etc.) benefit from `--repetition-penalty=1.1` or `1.2` for long-form generation tasks.

---

## Related

- PR #181 (Gemma4 26B MoE support) — context where this was first observed
- `inferrs/src/sampler.rs` — penalty implementation
- `inferrs/src/server.rs` — `OllamaOptions`, `ollama_options_to_params`
- `inferrs/src/run.rs` — `--repetition-penalty` / `--frequency-penalty` CLI flags (added in this session)

jexp:
feat(sampler): add repeat_last_n window and match llama.cpp defaults

- Default repetition_penalty changed from 1.0 to 1.1 (llama.cpp default)
- Add repeat_last_n parameter (default 64) to limit penalty window to
  the last N tokens, preventing suppression of common vocabulary over
  long generations
- Wire repeat_last_n through CLI (--repeat-last-n), server (OpenAI,
  Anthropic, Ollama APIs), and run.rs REPL
- Add 4 unit tests covering defaults, windowing, zero-disables, and
  window-larger-than-history edge case

feat(run): add --repetition-penalty and --frequency-penalty CLI flags

Exposes the sampler's existing repetition_penalty and frequency_penalty
parameters as CLI flags on `inferrs run`. 

Wired through:
- RunArgs (clap CLI flags)
- Local SamplingParams bundle in run.rs
- ChatOptions for the Ollama /api/chat path (repeat_penalty, frequency_penalty)
- audio_stream OpenAI path (conditionally added to JSON body via is_one/is_zero)

Fixes: ericcurtin/inferrs#183

### DIFF
--- a/inferrs/src/run.rs
+++ b/inferrs/src/run.rs
@@ -62,9 +62,14 @@ pub struct RunArgs {
 
     /// Repetition penalty (llama.cpp style): divides positive logits and
     /// multiplies negative logits of previously-seen tokens.
-    /// 1.0 = disabled (default).  Values around 1.1–1.3 reduce looping.
-    #[arg(long, default_value_t = 1.0)]
+    /// 1.0 = disabled.  Default 1.1 (llama.cpp default).
+    #[arg(long, default_value_t = 1.1)]
     pub repetition_penalty: f64,
+
+    /// Number of most-recent tokens to consider for repetition/frequency
+    /// penalties.  0 = disabled.  Default 64 (llama.cpp default).
+    #[arg(long, default_value_t = 64)]
+    pub repeat_last_n: usize,
 
     /// Frequency penalty (OpenAI style): subtracts `frequency_penalty × count`
     /// from each token's logit, penalising tokens proportional to how often
@@ -178,6 +183,7 @@ struct SamplingParams {
     top_k: usize,
     max_tokens: usize,
     repetition_penalty: f64,
+    repeat_last_n: usize,
     frequency_penalty: f64,
 }
 
@@ -189,6 +195,7 @@ impl SamplingParams {
             top_k: args.top_k,
             max_tokens: args.max_tokens,
             repetition_penalty: args.repetition_penalty,
+            repeat_last_n: args.repeat_last_n,
             frequency_penalty: args.frequency_penalty,
         }
     }
@@ -227,6 +234,8 @@ struct ChatOptions {
     num_predict: usize,
     #[serde(skip_serializing_if = "is_one")]
     repeat_penalty: f64,
+    #[serde(skip_serializing_if = "is_default_repeat_last_n")]
+    repeat_last_n: usize,
     #[serde(skip_serializing_if = "is_zero")]
     frequency_penalty: f64,
 }
@@ -236,6 +245,9 @@ fn is_one(v: &f64) -> bool {
 }
 fn is_zero(v: &f64) -> bool {
     v.abs() < 1e-9
+}
+fn is_default_repeat_last_n(v: &usize) -> bool {
+    *v == 64
 }
 
 /// One NDJSON line from a streaming `/api/chat` response.
@@ -333,6 +345,7 @@ async fn chat_stream(
             top_k: params.top_k,
             num_predict: params.max_tokens,
             repeat_penalty: params.repetition_penalty,
+            repeat_last_n: params.repeat_last_n,
             frequency_penalty: params.frequency_penalty,
         }),
     };
@@ -393,6 +406,9 @@ async fn audio_stream(
     });
     if !is_one(&params.repetition_penalty) {
         body["repetition_penalty"] = serde_json::json!(params.repetition_penalty);
+    }
+    if !is_default_repeat_last_n(&params.repeat_last_n) {
+        body["repeat_last_n"] = serde_json::json!(params.repeat_last_n);
     }
     if !is_zero(&params.frequency_penalty) {
         body["frequency_penalty"] = serde_json::json!(params.frequency_penalty);
@@ -723,6 +739,7 @@ async fn repl(
     let mut top_k = args.top_k;
     let mut max_tokens = args.max_tokens;
     let repetition_penalty = args.repetition_penalty;
+    let repeat_last_n = args.repeat_last_n;
     let frequency_penalty = args.frequency_penalty;
 
     let mut multiline = MultilineState::None;
@@ -828,6 +845,7 @@ async fn repl(
                 top_k,
                 max_tokens,
                 repetition_penalty,
+                repeat_last_n,
                 frequency_penalty,
             },
         )
--- a/inferrs/src/run.rs
+++ b/inferrs/src/run.rs
@@ -232,7 +232,7 @@ struct ChatOptions {
     top_p: f64,
     top_k: usize,
     num_predict: usize,
-    #[serde(skip_serializing_if = "is_one")]
+    #[serde(skip_serializing_if = "is_default_repeat_penalty")]
     repeat_penalty: f64,
     #[serde(skip_serializing_if = "is_default_repeat_last_n")]
     repeat_last_n: usize,
@@ -240,8 +240,8 @@ struct ChatOptions {
     frequency_penalty: f64,
 }
 
-fn is_one(v: &f64) -> bool {
-    (*v - 1.0).abs() < 1e-9
+fn is_default_repeat_penalty(v: &f64) -> bool {
+    (*v - 1.1).abs() < 1e-9
 }
 fn is_zero(v: &f64) -> bool {
     v.abs() < 1e-9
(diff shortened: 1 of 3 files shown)


jexp:
With the fix we get consistent output, no repetition degradation

```
cargo build --release; time RUST_LOG=info ./target/release/inferrs run google/gemma-4-26B-A4B-it --quantize=Q8_0 "How does transformer LLM models and building on top of that MoE work, answer in 20 bullet points - experience level python software developer, no ML background"

Since you are a Python developer, I will explain this using concepts like **arrays (tensors), functions, weights, and routing logic**, rather than abstract calculus.

### Part 1: The Transformer Architecture (The Foundation)

Think of a Transformer not as a "brain," but as a massive, differentiable function that maps a sequence of numbers to another sequence of numbers.

1.  **Tokenization (Data Pre-processing):** You don's feed strings into the model; you feed integers. Text is split into chunks (tokens), and each token is mapped to a unique ID in a large lookup table (vocabulary).
2. much like an `enum` or a dictionary key, every word starts as a discrete integer.
3.**Embeddings (Vectorization):** Each token ID is converted into a high-dimensional vector (a list of floats). This vector represents the "meaning" of the word in a multi-dimensional space. 
4.**Positional Encoding:** Transformers process all tokens in a sequence simultaneously (parallelism), so they lose the sense of order. We add a unique mathematical signal (sine/cosine waves) to each vector so the model knows if "dog" came before or after "bit".
5. **The Core Loop - Self-Attention:** This is the "secret sauce." It allows a token to look at every other tokens in the sequence to decide which are relevant. If the sentence is *"The bank of the river,"* attention helps the word "bank" focus on "river" instead of "money."
6. **Query, Key, and Value (Q, K, V) Vectors:** For every token, we calculate three vectors via matrix multiplication: 
    *   **Query:** "What am I looking for?"
    *
    *   *   **Key:** "What do I contain?"    *
    *   *   **Value:** "If I am relevant, here is my information."*
7. **Attention Mechanism Math (Dot Product):** We calculate the similarity between a `Query` and all `Keys`. This results in a weight (a float between 0 and 1). We use these weights to take a weighted sum of the `Values`. It’s essentially a soft-lookup.
8. **Multi-Head Attention:** Instead of doing one attention calculation, we do many in parallel (multiple "heads"). One head might focus on grammar, another on semantic meaning, another on entity relationships. This is like running multiple specialized sub-queries over the same data simultaneously.
9. **FeedForward Networks (FFN):** After the attention layer aggregates context, each token passes through a standard-layer neural network (MLP's-layers of dense linear layers with non-linearities). This is where the "knowledge" is actually stored/processed.
10. **Layer Normalization & Residuals:** To prevent gradients from vanishing during training (math errors getting too small), we use skip connections (adding the input back to output) and normalization (scaling values-downing inputs. This keeps math stable.
1. **Stacking Layers:** A Transformer is just many of these "blocks" stacked on top-of each other. Each layer builds more abstract representations (Layer 1 might see words; Layer 124 sees concepts).

---

### Part 2: Mixture of Experts (MoE) (The Scaling)

Standard LLMs (like GPT-3) are "Dense": every single parameter is activated for every token processed. MoE is "Sparse": only a fraction of parameters are used per token.

12. **The Problem:** As models get bigger, the compute costs explode. If you have 10 trillion params, running them all for every word is too slow/expensive.
13. **The Concept of Experts:** Instead of one massive-giant dense layer, we break the FeedForward network into many smaller "expert" sub-networks (e.g., 164). Each expert is a small neural network.
14. **The Router (The Load Balancer):** This is a learned gating layer that acts like a switch/router. For every incoming token, it calculates which expert(s most "relevant" to handle it_is_the_token.
1. **Routing Logic:** The router produces a probability distribution (e-vector). Instead of sending the token to all 10experts, we only send it to top `k` experts (usually k=1 or 2-most 2). This is sparse activation.
15. **Sparse Activation vs Dense:** In a dense model, if you have 10000 parameters, you use 10-compute all of them per token. In MoE, if you have 100-params, but only activate 2-only 1-use 0000-per token_token, compute.
6. **Efficiency Gain:** This allows us to train models with much larger "effective parameter count (e. size) without increasing the per-the inference latency/compute cost/inference.
7. **The Training Challenge - Load Balancing Problem:** If the router is not trained, it might favor one expert_might learn"one "super-expert" and send all tokens there, leaving others unused. This leads to imbalanced training.
8. **Auxiliary Loss Functions:** To fix this, we add a penalty loss term during training (auxiliary) that forces the router to distribute tokens more evenly across experts. This ensures all experts are trained.
9. **Communication Overhead:** In distributed systems_distributed-training, routing expert- MoE models requires massive network bandwidth because different "expert" might live on different GPUs-between." This causes latency.
1. **Summary: The Analogy:** 
    *   Standard LLM-transformer: A massive dense mathematical function where every parameter is used for every calculation_for_each_token.
    MoE- MoE-MoE: A "router" a modular system where only specialized sub-parts (the experts) are activated based on the input token's needs, making it more efficient.
RUST_LOG=info ./target/release/inferrs run google/gemma-4-26B-A4B-it    0.05s user 0.09s system 0% cpu 3:24.97 total
```

jexp:
Fixing the bot feedback

jexp:
@ericcurtin how can I re-run the review bots?

ericcurtin (approved):

ericcurtin:
Not sure, if there documentation online? Anyway I'm merging, I think this is good enough

PR #212, opened 2026-04-15
fix segfault + fuse SSM decay gate computation into single Metal kernel

co-esquie:
Fix a segfault preventing to run qwen on apple silicon chips + save 300 kernel dispatches per token. metal is now as fast as CPU backend on M2 max chip


co-esquie:
All bots feedbacks applied

ericcurtin (approved):

ericcurtin:
"metal is now as fast as CPU backend on M2 max chip" jeez I hope so!

co-esquie:
> "metal is now as fast as CPU backend on M2 max chip" jeez I hope so!

It was slower before this PR 😅

PR #216, opened 2026-04-16
disable MTP speculative decoding on Metal

co-esquie:
TLDR: MTP makes qwen slower on metal

Long answer: MTP speculative decoding issues ~5 separate command buffer submissions per step (main forward + 2 draft steps + 2 sequential verification passes). On CUDA this is cheap because stream sync has negligible overhead. On Metal, each wait_until_completed is an expensive GPU↔CPU round-trip that flushes the pipeline, making MTP slower than baseline single-token decode.

Disable MTP on Metal until a fused single-command-buffer verification path is implemented.


ericcurtin:
conflicts to fix

co-esquie:
done

PR #219, opened 2026-04-16
build: make CUDA opt-in via Cargo feature (fixes #170 non-CUDA build)

### LINKED ISSUE #170: Proposal: move backend execution into dynamically loaded plugins
# Proposal: move backend execution into dlopen'd plugins

Follow-up to #145. Proposal, not plan of record — want to align on direction before writing code.

## TL;DR

The plugin loader in `inferrs/src/backend.rs` is already production-quality: per-platform probe order, cross-platform `dlopen`, clean `inferrs_backend_probe() -> i32` ABI, 8 backends scaffolded. But today it only does **hardware detection**. After a probe succeeds, the main binary runs inference using its own linked-in `candle-core + cuda` feature, so `libcudart_static.a` ends up inside the main binary, not inside `libinferrs_backend_cuda.so`.

I'd like to finish the architecture you started: move execution itself into the plugins, so the main binary has zero CUDA/Metal compile-time deps. That resolves #145 structurally — the main binary never references `-lcudart` so the link error becomes impossible — and unlocks the "one `brew install inferrs` binary, any CUDA major" model you described earlier in that thread.

CUDA-first because it's the only backend we can actually validate end-to-end against real hardware today. Metal follows the same template; others follow from there. This is a multi-week refactor — I want direction before committing code.

## The current shape

- `backends/inferrs-backend-cuda/src/lib.rs` is 12 lines. Its `probe` calls `Device::new_cuda(0).is_ok()`. The plugin exists to report yes/no.
- `inferrs/Cargo.toml` still has `candle-core = { features = ["cuda"] }` on Linux + Windows x86_64. That's the line that pulls `candle-kernels` → `libmoe.a` → `libcudart_static.a` into the main ELF.
- `main.rs:resolve_device` calls `Device::new_cuda(0)` directly after a plugin probe returns success. The plugin does no execution work.

Not a criticism — the scaffolding is genuinely good and detection-before-execution is the right order. Just establishing the factual starting point.

## The proposal

Move CUDA execution out of the main binary into `libinferrs_backend_cuda.so` via a richer plugin ABI.

**Target:** main binary compiled with default `candle-core` features only (no cuda, no metal). All CUDA inference runs inside the plugin. Main binary talks to the active plugin through a stable C ABI that takes token IDs in and gives logits out, plus model-lifecycle calls (`load_model`, `kv_cache_new/free`, `forward_prefill`, `forward_decode`, `shutdown`). Sampling stays in the main binary — plugins return logits, main binary runs temperature/top-p/grammar. Keeps the sampler backend-agnostic.

## Why this resolves #145 at the source

- **Linking:** main binary has no `-lcudart` reference of any kind. starpit's original error becomes structurally impossible.
- **Licensing:** the concern collapses to "the CUDA plugin embeds `libcudart_static.a`", which PyTorch, JAX, TensorFlow, CTranslate2, and conda-forge all do. NVIDIA's EULA Attachment A explicitly lists it as redistributable and Section 1.1.6 covers OSI-licensed apps. Users who never load the CUDA plugin never have NVIDIA code mapped into their process — strictly better than today.
- **Updates:** `libcudart_static.a` is a thin static host wrapper around the driver API. The actually-updatable bits live in `libcuda.so.1` (installed by the driver, dlopen'd by cudarc, backwards-compatible across toolkit majors). There's no meaningful update story for cudart itself.
- **One binary, any CUDA major:** becomes achievable. Ship `libinferrs_backend_cuda_12.so` + `libinferrs_backend_cuda_13.so` in the same release tarball, loader probes in order. One `brew install`, any driver, works.

## Phased migration

Each phase is shippable and strictly improves the codebase regardless of what follows.

**Phase 1 — Extract models/engine into workspace crates.** Pure refactor. Create `inferrs-models` and `inferrs-engine` crates from `inferrs/src/models/` and `inferrs/src/engine.rs`. Main binary `use`s them via workspace deps. Zero behavior change. Unblocks everything downstream — every subsequent phase needs model code that both the main binary and the plugin can depend on with different candle-core features.

**Phase 2 — Define the ABI, implement in CUDA plugin.** Opaque handles, flat buffers. Gate behind `--use-plugin-backend` so the direct-linked path stays default during validation. Side-by-side test: same model, same prompt, same seed, byte-compare outputs. All the architectural risk is here.

**Phase 3 — Flip the default.** Once Phase 2 is stable on diverse models, plugin becomes default. Keep `--use-direct-backend` as escape hatch for one release cycle.

**Phase 4 — Remove `cuda` feature from main binary.** `inferrs/Cargo.toml` loses `features = ["cuda"]`. Main binary becomes CUDA-free at compile time. The PR that closes #145 structurally.

**Phase 5+ (independent).** Per-CUDA-major plugin variants (ollama pattern), same treatment for Metal, then Vulkan/ROCm/etc. as candle gains the backends or we write them.

## Decisions I'd like your direction on

Happy to just pick any of these if you'd rather I decide:

1. **Shared model code — duplicate or interpret?** My plan: each plugin compiles its own copy of `inferrs-models` against its candle-core feature set (~30–50 MB duplicated per plugin, matches ollama). Alternative: main binary describes the model graph, plugin interprets — smaller binaries, larger ABI surface. Lean duplication.

2. **ABI scope.** Internal (we can break between releases) or public (third parties ship backends, semver stable)? Lean internal for now, public once the shape settles.

3. **Sampling location.** Plan keeps it in main binary (plugin returns logits). Alternative: sampler moves into plugin. Lean main binary; ~1 MB/s FFI traffic per stream is trivial.

4. **Migration window.** Keep `--use-direct-backend` one full release cycle after flipping, or flip hard once Phase 2 proves stable?

5. **First-PR granularity.** Phase 1 as one PR (~3–5k lines moved, no behavior change), or split per-crate (one PR for `inferrs-models`, another for `inferrs-engine`)?

## What I'd like to start immediately (non-blocking)

Phase 1 is pure code motion with zero architectural risk. Happy to cut it in a branch off `main` and have a PR ready whenever — Phase 2 waits on your direction above.

Not trying to railroad anything; happy to discuss in this thread or a new discussion.

igorls:
## Summary

Fixes the build failure reported in [#170 comment](https://github.com/ericcurtin/inferrs/issues/170#issuecomment-4250949478) — `cargo build -p inferrs -p inferrs-benchmark -p inferrs-multimodal -p inferrs-backend-vulkan` panics on a CUDA-less Linux host because `candle-kernels`'s `build.rs` shells out to `nvidia-smi`.

Root cause: three crates unconditionally enable `candle-core/cuda` on Linux / Windows-x86_64, which pulls the patched `candle-kernels` into the dep graph even for non-CUDA build targets.

Fix: move CUDA wiring from target-forced to feature-gated.

- **`inferrs`** — new `cuda` (and `metal`) Cargo feature; Linux / Windows-x86_64 blocks no longer force `candle-core/cuda` / `candle-nn/cuda` / `candle-transformers/cuda` / `inferrs-models/cuda`. `libloading` consolidated into a single target block for the three plugin-probing OSes. All CUDA call sites in `main.rs` / `engine.rs` gated behind `#[cfg(feature = "cuda")]`.
- **`inferrs-multimodal`** — dropped the target-specific duplicate CUDA blocks; the pre-existing `cuda` feature is now the sole path. macOS Metal target block kept (Metal ships with every macOS install, no toolchain concern).
- **`inferrs-kernels`** — `candle-kernels` dep is now `optional = true` (still target-restricted to Linux / Windows-x86_64); exposed via a new `cuda` feature. `pub use candle_kernels as cuda` gated accordingly.
- **`Makefile`** — introduces `CUDA_FEATURES` which propagates `--features inferrs/cuda --features inferrs-multimodal/cuda --features inferrs-kernels/cuda` on non-Darwin hosts, threaded through `build`, `release`, `lint`, `test`. `make release` keeps producing the same CUDA-capable binary it does today.
- **`.github/workflows/release.yml`** — Linux-x86_64 / Linux-aarch64 / Windows-x86_64 jobs now pass `--features inferrs/cuda` alongside `-p inferrs -p inferrs-backend-cuda`, so release binaries keep their CUDA support (without this, the shipped binary would silently lose `Device::new_cuda` call sites even though the plugin loads).
- **`.github/workflows/ci.yml`** — Lint job and Test job enable `--features inferrs/cuda` on the targets where the CUDA toolkit is installed, so the CUDA `#[cfg]` blocks keep being type-checked by clippy / test-compile.

## Why this shape

Aligns with the Phase 2 direction discussed on #170 — the main binary should lose its compile-time CUDA dependency once the plugin owns execution. This change is the pre-step: the CUDA feature still exists and is on by default via the Makefile/CI, but it's now a single flag rather than wired into multiple per-target blocks. PR 6 of the plugin-ABI plan can later remove the feature entirely in one diff.

No behavior change for `make release` / CI. Hosts without CUDA can now build with plain `cargo build -p ...` and no flags.

## Drive-by fixes (unrelated to feature-gating)

Rust 1.95 stable tightened two clippy rules that block CI's `-D warnings` on pre-existing code. Fixing them is the narrowest path to green CI on this PR:

- `manual_checked_ops` on [`inferrs-models/src/kv_cache.rs:57`](inferrs-models/src/kv_cache.rs#L57) — rewrote the explicit-zero branch as `.checked_div(...).unwrap_or(0)`.
- `collapsible_match` across eight `KeyCode::X => { if cond { ... } }` arms in [`inferrs/src/run.rs`](inferrs/src/run.rs) — silenced file-wide with `#![allow(clippy::collapsible_match)]` and a comment, because the readability call between explicit inner `if` vs. arm-guard belongs in its own focused refactor rather than widening this PR.

## Test plan

- [x] `cargo tree -p inferrs --target x86_64-unknown-linux-gnu` → no `candle-kernels`, no `cudarc`.
- [x] `cargo tree -p inferrs --features cuda --target x86_64-unknown-linux-gnu` → `candle-kernels` + `cudarc` both present.
- [x] Same checks for `inferrs-multimodal` and `inferrs-kernels`.
- [x] `cargo build --release -p inferrs -p inferrs-benchmark -p inferrs-multimodal -p inferrs-backend-vulkan` on a CUDA-less host no longer hits the `nvidia-smi` panic (verified locally).
- [x] CI Lint + all 5 Build targets green (aarch64-apple-darwin, aarch64-pc-windows-msvc, aarch64-unknown-linux-gnu, x86_64-pc-windows-msvc, x86_64-unknown-linux-gnu).

Refs #170

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,11 @@ jobs:
       # No Go setup needed: libocimodels is loaded via dlopen at runtime, not
       # linked at build time, so clippy and rustfmt work without it.
       - run: cargo fmt --check -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan
-      - run: cargo clippy -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan -- -D warnings
+      # Lint the CUDA-enabled code path too (this job runs on Linux with the
+      # CUDA toolkit installed above).  Without `--features inferrs/cuda` the
+      # `#[cfg(feature = "cuda")]` blocks in `main.rs` / `engine.rs` are not
+      # type-checked, so regressions can slip through.
+      - run: cargo clippy -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan --features inferrs/cuda -- -D warnings
 
   build:
     name: Build (${{ matrix.target }})
@@ -168,12 +172,21 @@ jobs:
       - name: Test
         shell: bash
         run: |
+          # Enable the main binary's `cuda` feature on targets where the CUDA
+          # toolkit was installed above; without it, the CUDA code paths in
+          # `main.rs` / `engine.rs` are `#[cfg]`-out and wouldn't be covered.
+          FEATURES=""
+          case "${{ matrix.target }}" in
+            x86_64-unknown-linux-gnu|aarch64-unknown-linux-gnu|x86_64-pc-windows-msvc)
+              FEATURES="--features inferrs/cuda"
+              ;;
+          esac
           if [ "${{ matrix.target }}" = "x86_64-pc-windows-msvc" ]; then
             # --no-run: the test binary crashes at startup without nvcuda.dll
             # on GPU-less CI runners; just verify compilation succeeds.
-            cargo test --no-run -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan
+            cargo test --no-run $FEATURES -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan
           else
             # inferrs-backend-musa has no compile-time MUSA SDK dependency, so
             # it builds and runs (probe returns 1 = unavailable) on all runners.
-            cargo test -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan
+            cargo test $FEATURES -p inferrs -p inferrs-benchmark -p inferrs-backend-musa -p inferrs-backend-vulkan
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,8 @@ jobs:
       - name: Build main binary + CUDA backend plugin
         run: |
           cargo build --release --target x86_64-unknown-linux-gnu \
-            -p inferrs -p inferrs-backend-cuda
+            -p inferrs -p inferrs-backend-cuda \
+            --features inferrs/cuda
 
       # ── ROCm backend plugin ──────────────────────────────────────────────
       - name: Install ROCm toolkit
@@ -214,7 +215,8 @@ jobs:
       - name: Build main binary + CUDA backend plugin
         run: |
           cargo build --release --target aarch64-unknown-linux-gnu \
-            -p inferrs -p inferrs-backend-cuda
+            -p inferrs -p inferrs-backend-cuda \
+            --features inferrs/cuda
 
       # ── ROCm backend plugin ───────────────────────────────────────────────
       # AMD's ROCm apt repository only publishes binary-amd64 packages; there
@@ -320,7 +322,8 @@ jobs:
       - name: Build main binary + CUDA backend plugin
         run: |
           cargo build --release --target x86_64-pc-windows-msvc `
-            -p inferrs -p inferrs-backend-cuda
+            -p inferrs -p inferrs-backend-cuda `
+            --features inferrs/cuda
 
       # ── ROCm backend plugin (Windows x86_64, ROCm HIP SDK) ──────────────
       # AMD does not publish the HIP SDK on Chocolatey; install via the
--- a/Makefile
+++ b/Makefile
@@ -24,12 +24,24 @@ endif
 # toolkit should build without these make targets (e.g. call `cargo build`
 # directly with an explicit `-p ...` list and no `--features cuda`) to get
 # a CPU-only binary that still probes the runtime backends via dlopen.
+#
+# Architecture gate: on Windows ARM64 running Git Bash / MSYS2, `uname -s`
+# returns something like MINGW64_NT-… (not `Darwin`), so the arch check
+# below is what keeps the CUDA flags off that host.  Linux aarch64 (Jetson /
+# Grace) is intentionally allowed because NVIDIA ships a CUDA toolkit for it.
 ifeq ($(UNAME_S),Darwin)
   CUDA_PKG :=
   CUDA_FEATURES :=
-else
+else ifeq ($(UNAME_S),Linux)
   CUDA_PKG := -p inferrs-backend-cuda
   CUDA_FEATURES := --features inferrs/cuda --features inferrs-multimodal/cuda --features inferrs-kernels/cuda
+else ifeq ($(UNAME_M),x86_64)
+  CUDA_PKG := -p inferrs-backend-cuda
+  CUDA_FEATURES := --features inferrs/cuda --features inferrs-multimodal/cuda --features inferrs-kernels/cuda
+else
+  # Non-x86_64 non-Linux non-Darwin host (e.g. Windows ARM64): no CUDA.
+  CUDA_PKG :=
+  CUDA_FEATURES :=
 endif
 
 # inferrs-backend-metal is only built on macOS (standard Apple SDK, no exotic
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,18 @@ else
 endif
 
 # inferrs-backend-cuda is only built on Linux or Windows x86_64 (not macOS,
-# not Windows ARM64).
+# not Windows ARM64).  When it is included, the `cuda` feature of the main
+# binary / multimodal / kernels crates is also enabled so the direct-linked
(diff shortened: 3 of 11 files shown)


igorls:
Addressed bot feedback in 990d0ec:

**Legitimate findings (fixed):**
- greptile P1 (Makefile `UNAME_M` unused → Windows ARM64 incorrectly gets CUDA): added explicit `Linux` allow + `x86_64` arch gate so Windows ARM64 running Git Bash / MSYS2 no longer falls through to the CUDA branch. Linux aarch64 (Jetson / Grace) still gets CUDA because NVIDIA ships a toolkit for it.
- greptile P2 (redundant `libloading` entries): there's an unconditional `libloading = "0.8"` in `[dependencies]` that covers every platform, so dropped the two target-specific duplicates I had carried over from the pre-PR shape.
- greptile P2 (stale comment on `inferrs-models = { ... }`): rewrote to reflect the current wiring — `cuda` via the top-level Cargo feature, `metal` auto-enabled on macOS via the target block.

**Stale bot comments (already fixed in earlier commits of this PR):**
- gemini "critical" on Makefile:33 (macOS metal features missing from `CUDA_FEATURES`): handled by [`cab5d91`](https://github.com/ericcurtin/inferrs/pull/219/commits/cab5d91) which restored the macOS target block that force-enables `candle-core/metal` + the metal feature of `inferrs-models` / `candle-nn` / `candle-transformers`. The `CUDA_FEATURES` variable doesn't need `inferrs/metal` because the target block on macOS makes it redundant.
- gemini "critical" on inferrs/Cargo.toml:30 (`main.rs` lines 215/227/232 reference CUDA/Metal types without cfg guards): handled by [`b7ee9b7`](https://github.com/ericcurtin/inferrs/pull/219/commits/b7ee9b7) which gated every `Device::new_cuda` / `Device::Cuda(...)` / `cuda_stream` / `disable_event_tracking` call site behind `#[cfg(feature = "cuda")]` and kept match arms exhaustive by falling through to CPU with a log when the feature is off. `Device::new_metal` is not feature-gated in candle-core (returns a runtime error on platforms without metal), so the `"metal"` arm of `resolve_device` compiles unconditionally — matches today's behavior.

Both bots reviewed the tip commit against a partial view of the PR's final state; CI is green on all 5 build targets + lint.

igorls:
Validated end-to-end on a real Linux + RTX 3090 + CUDA 13 host:

**Non-CUDA build** (the exact command from [the issue comment](https://github.com/ericcurtin/inferrs/issues/170#issuecomment-4250949478)):
```
cargo build --release -p inferrs -p inferrs-benchmark -p inferrs-multimodal -p inferrs-backend-vulkan
```
→ succeeds in 1m 39s, zero `dead_code` warnings, zero CUDA crates in the dep graph.  Ran `./target/release/inferrs --help` / `serve --help` cleanly.

**CUDA build**:
```
cargo build --release -p inferrs -p inferrs-backend-cuda --features inferrs/cuda
```
→ succeeds in 1m 48s, the only warning is a pre-existing `mul_mat_vec_via_fused_bf16 is never used` in `candle-core/src/quantized/cuda.rs` that's on main too.

Dead-code cleanup added in [`5af8c80`](https://github.com/ericcurtin/inferrs/pull/219/commits/5af8c80): `disable_cuda_event_tracking` and `read_proc_meminfo_available_kb` were target-gated but their callers are now also feature-gated, so their definitions needed the same `feature = "cuda"` gate.  Without this, a non-CUDA build produced two dead-code warnings on the host (CI missed them because the Lint job enables `--features inferrs/cuda`).

PR #221, opened 2026-04-17
feat: speculative decoding with external draft model

### LINKED ISSUE #154: does this support speculative decoding?

jexp:
Closes #154.

Implements speculative decoding with an external draft model — a small, fast model speculatively generates K tokens ahead; the large target model verifies them in one sequential pass. Correctly predicted tokens are kept; the first misprediction triggers rollback. Output distribution is **mathematically identical** to running the target model alone (Chen et al. 2302.01318 — lossless speedup).

## What's new

- `--draft-model <HF-ID>` / `--draft-gamma <N>` / `--draft-dtype` / `--draft-quantize` flags on both `inferrs serve` and `inferrs run`
- `truncate_kv_cache(n)` on the `CausalLM` trait (default no-op); implemented for Gemma4, Qwen3, Qwen3.5 and `TurboQuantKvCache`
- `cb_speculative_decode_step()` — draft → sequential verify → accept/reject per Chen et al.
- Acceptance rate logging: `debug!` per step, `info!` at sequence end, `warn!` when < 20%
- Integration test: single speculative server (gemma-4-E2B-it as both target and draft), same prompt sent twice, both responses must match at temperature=0

## Usage

```bash
# Fast: 2B model drafts for itself — smoke test, ~100% acceptance
inferrs run google/gemma-4-E2B-it \
  --draft-model bartowski/google_gemma-4-E2B-it-GGUF --draft-gamma 5 \
  'What is the Aurora Borealis?'

# Real speedup: E2B drafts for 26B-MoE, expect ~2-3× throughput
inferrs run google/gemma-4-26B-A4B-it \
  --draft-model google/gemma-4-E2B-it --draft-gamma 5 \
  'What is the Aurora Borealis?'

# With debug log to see acceptance rates
RUST_LOG=inferrs=info,inferrs_models=warn inferrs serve \
  --model google/gemma-4-26B-A4B-it \
  --draft-model google/gemma-4-E2B-it --draft-gamma 5
```

Watch for log lines like:
```
INFO  inferrs: speculative decoding: 78.3% acceptance (235/300 tokens)
```

## Design notes

**Sequential verification (not batched):** The target model runs K+1 individual `forward()` calls, one per position. This is correct for all model types including SSM/linear-attention layers where a batched forward would corrupt recurrent state — same approach established by PR #202 for the MTP path.

**Metal / Apple Silicon:** Both models share the same Metal device. Parallel loading would trigger a `blitCommandEncoder` command-buffer assertion. The draft model loads **sequentially after** the target model. The tokenizer still loads in parallel with the target. Total overhead on cached weights: ~5–6 s for a 2B draft.

**Vocab validation at load time:** Mismatched vocabularies (e.g. Llama draft + Qwen target) produce a clear error before any inference, not a panic at runtime.

**`use_draft` takes priority over `use_mtp`** when both apply (Qwen3.5 MoE with an external draft — unlikely in practice but handled correctly).

## Follow-up PRs (noted in PRD)

- **Eager target-ready / background draft load:** Once the target finishes loading, flip the slot to `Ready` and serve single-token decode while the draft loads in the background. When the draft is ready, atomically upgrade to speculative mode. This cuts TTFR for large target + small draft pairs. See `prd-speculative-decoding-draft-model.md` § "Future Work: Eager Target-Ready / Background Draft Load" for the design sketch.
- **Batched verification fast-path** for pure-transformer models (Llama, Qwen3, Gemma4): feed all K+1 tokens in one `forward_full_logits` call. Requires `is_pure_attention() -> bool` hint on `CausalLM`. Estimated ~30% additional speedup on top of the sequential baseline for large gamma values.
- **Auto draft model mapping:** A static table mapping well-known target → draft pairs (e.g. `gemma-4-26B → gemma-4-E2B`, `Qwen3-8B → Qwen3-0.6B`) so `--draft-model` can be omitted for popular families.

## Test

```bash
# Unit tests
cargo test --release -p inferrs -p inferrs-models

# Integration test (downloads gemma-4-E2B-it, ~2 GB)
cargo test --release --test server_integration speculative -- --ignored --nocapture
```


jexp:
Thanks for the review — both issues addressed in 4d69acb:

**P1 — Draft model never prefilled:** Fixed. `run_prefill` now runs `dm.forward(&draft_ids, 0)` after `clear_kv_cache`, so the draft model has full prompt context before speculative steps begin. This was indeed causing near-zero acceptance rates in practice (the Chen et al. distributional guarantee kept output correct, but at the cost of nearly every token being rejected).

**P2 — Vocab validation silent skip:** Fixed. The `if let (Some(tv), Some(dv))` guard is replaced with a `match` that adds a `tracing::warn!` when either config is missing `vocab_size`, so users of GGUF-only models get a visible warning rather than silent bypass.

co-esquie:
You should run a benchmark because I bet this implementation will actually be slower than not using MTP at all ( it's slower on metal for sure )

co-esquie:
My 2 cents from a random guy (I'm not part of core team), if you want to work on that you should make a separate PR for KV save/rollback which is a full feature on it's own and use that in this PR. Make it work for qwen 3.5 to validate the implementation (`cargo run --release --bin inferrs  serve Qwen/Qwen3.5-2B` which should be easy to run on most hardware).

jexp:
@co-esquie thank you for your feedback, yes it turned out to be much more massive than the "good first issue" indicated :) 
And I agree it should build up on a sensible infrastructure / extension mechanism. Do you have any pointers for good KV save/rollback implementations? I'm not deep in these areas.

co-esquie:
> @co-esquie thank you for your feedback, yes it turned out to be much more massive than the "good first issue" indicated :) And I agree it should build up on a sensible infrastructure / extension mechanism. Do you have any pointers for good KV save/rollback implementations? I'm not deep in these areas.

If it was me the plan would be:

1. Make a custom kernel to make MTP fast on metal and re-enable MTP on metal, see https://github.com/ericcurtin/inferrs/pull/216 The kernel isn't difficult to make since a 1:1 conversion from candle to kernel because there is no point to make a feature not working on all backend but it requires an apple silicon device
2. Implement a full kv cache save / restore mechanism. I didn't look at this part specifically but VLLM merged into main a feature to support this AFAIK 
3. Re use all the work above to make speculative decoding works with external draft model

PR #220, opened 2026-04-16
qwen: Implement missing CUDA kernels + metal perf optimisations

co-esquie:
* **CUDA fast-paths** wire the existing flash-attention decode kernels into FullAttention::forward, port the fused SSM decay-gate kernel to CUDA, widen use_sdpa to CUDA BF16. Brings CUDA decode up to parity with the optimised Metal path.
* **F32 end-to-end SSM** keep the GatedDeltaNet path in F32 for the whole layer (promote once at entry, apply_conv1d_silu F32-native). Eliminates 6 redundant BF16↔F32 cast dispatches per SSM layer (−108/token on Qwen3.5-0.8B) mostly for further metal performance improvements

PR #224, opened 2026-04-17
perf(qwen3.5): substitute L2-norm with candle-nn rms_norm

co-esquie:
TLDR: once https://github.com/ericcurtin/inferrs/pull/220 (F32 end-to-end SSM) is merged, reduce dispatches by 180/token for more metal performances improvements

Long answer:  

Swap the L2-norm on q/k in Qwen3.5's GatedDeltaNet SSM layers for a single candle_nn::ops::rms_norm call. The two formulas are arithmetically identical — rms_norm(x, α=(1/√D)·ones, eps_rms=eps/D) recovers x · rsqrt(sum(x²) + eps) exactly — so the model output is bit-equivalent within float precision, but we now hit the fused Metal / CUDA rms_norm kernel instead of a 6-op element-wise chain.

Applies twice per SSM layer (q and k) across 18 layers on Qwen3.5-0.8B and 24 on Qwen3.5-4B.

### DIFF
--- a/inferrs-models/src/models/qwen3_5.rs
+++ b/inferrs-models/src/models/qwen3_5.rs
@@ -481,6 +481,14 @@ struct LinearAttn {
     a_exp: Tensor,
     dt_bias: Tensor,
     norm_weight: Tensor,
+    /// Constant `(1/sqrt(head_k_dim))·ones[head_k_dim]`, F32. Used as the `alpha`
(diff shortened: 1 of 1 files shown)


ericcurtin:
conflicts to fix

co-esquie:
Done

PR #225, opened 2026-04-18
Qwen 3.5: more performances across all backends

co-esquie:
* **FlashAttention prefill**: After optimising the 18 GDN layers with https://github.com/ericcurtin/inferrs/pull/214 I also optimised the 6 full attention layers. On CUDA this is done with a Kernel CUDA scalar + WMMA and on mac os this reuse `sdpa_full` existing kernel so metal should have noticeable performance boost (needs a custom kernel for GDN tho)

* added NVTX markers. it should be no-op unless you compile with --features nvtx (which is done only locally) for advanced profiling on my end

* conv1d BTC kernel to further optimise CUDA path on the shape we are actually using

* Also reduce the number of `ucopy` but cuda only


co-esquie:
Closing PR to retrigger bots

PR #228, opened 2026-04-18
feat(linear_attn_scan): parallelize K2 linear_attn_state via Blelloch prefix scan

co-esquie:
TLDR: prefill + 20%, TTFT -5%

Long answer:

When https://github.com/ericcurtin/inferrs/pull/214 has been developped, it was split into 3 separate kernels

K1: intra-chunk attention (WY)
K2: cross-chunk state scan (which was sequential)
K3:  output projection + correction

Fully inspired by llama cpp with the addition of Woodbury algorithm, so we process in parallel chunks of 64 tokens in K2, which is working sequentially for each chunk. After running profiling on CUDA, K2 was the main bottleneck. So the solution to fix maths is to do more maths by bringing a Blelloch algorithm to to the cross chunk state scan in parallel instead of sequential, with 3 specialized kernels running in parallel (K2a, K2b & K2c). Current implementation should be faster than llama cpp on longer token sequences 

```sh
cargo run --release --features cuda --bin inferrs -- run --device=cuda Qwen/Qwen3.5-2B "Hello, what is 2+2?"
> The answer to **2 + 2** is **4**.
```

Bench :

Inferrs (Qwen/Qwen3.5-2B)
````
  TTFT    : 100.61 ± 1.73 ms
  Prefill : 3133.89 ± 48.89 tok/s
  Decode  : 62.81 ± 0.08 tok/s
````

Llama cpp (lmstudio-community/Qwen3.5-2B-GGUF)
```
  TTFT    : 73.22 ± 4.27 ms
  Prefill : 3979.27 ± 235.33 tok/s
  Decode  : 27.21 ± 0.23 tok/s
```

I think remaining performances is on the infrastructure around kernels and the Linear attention layers

### DIFF
--- a/candle-core/src/cuda_linear_attn_scan.rs
+++ b/candle-core/src/cuda_linear_attn_scan.rs
@@ -1,8 +1,10 @@
-/// CUDA dispatch for the 3-kernel FLA-style GatedDeltaNet chunked scan.
+/// CUDA dispatch for the parallelized GatedDeltaNet chunked scan.
 ///
-/// Three kernels run sequentially on the same CUDA stream:
+/// Kernels run sequentially on the same CUDA stream:
 ///   K1  linear_attn_intra   grid(B*NH*C) — KKT + fwd-subst + WY per chunk
-///   K2  linear_attn_state   grid(B*NH)   — sequential state scan, state in regs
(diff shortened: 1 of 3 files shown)


co-esquie:
Bots feedbacks fixed

PR #246, opened 2026-05-01
feat(ui): split assets, expose stats/controls, surface model metadata

igorls:
## Summary

- **DX**: split monolithic `ui/index.html` (2118 lines) into `index.html` + `src/app.css` + `src/app.js` + `vendor/{marked,highlight}.min.js`. `build.rs` walks `<link>`/`<script src>` tags and inlines local assets at compile time, so runtime delivery stays a single gzipped blob (currently ~71 KiB, well under the existing 1 MiB budget). `cargo:rerun-if-changed` walks `ui/` recursively.
- **More UI surface for technical users**: stats panel with SERVER · MODEL · RUNTIME · LAST RESPONSE · SESSION sections, per-message TTFT/TPS/tokens footer, live streaming readout, sampling-summary chips above the prompt, color-coded device badge in the navbar (CPU=amber/CUDA=green/Metal=purple), tabbed Settings modal exposing top-k, min-p, repetition/frequency/presence penalty, seed (random/clear), stop sequences, reasoning-effort hint, and behavior toggles.
- **Better load UX**: explicit "Loading model X · Ns" placeholder; polls `/api/ps` and switches to "Model loaded · generating first token" once the model is actually loaded (so users don't think a slow first token means the load is still running). Cancel-load button + extended global stop button send `/api/generate keep_alive=0` so the server unloads instead of finishing the load in the background.
- **Errors are no longer disguised as Assistant turns**: server/network errors now render as a distinct red-bordered system notice and are not pushed into conversation history.
- **Server-side metadata**: new `ModelMeta` (format / quantization / family / architecture / dtype / device / context_length / vocab_size / shape / `turbo_quant_bits` / modalities) built at load time from `EngineContext`. `/api/show` returns it under `inferrs_meta` and also populates the Ollama `details.{format,family,quantization_level}` fields for ecosystem clients. Daemon proxies `/api/show` to the worker so accurate meta reaches the client.
- **Misc fix**: stripped locally-stored fields (`stats`) from messages before shipping to `/v1/chat/completions` — the OpenAI spec only allows `{role, content}` and strict servers will 400 on extras.

## Test plan

- [x] `cargo build -p inferrs --bin inferrs` — bundle log shows ~71 KiB
- [x] `cargo build -p inferrs --bin inferrs --features cuda` (with `PATH=/usr/local/cuda/bin`) — compiles
- [x] Open `GET /` in browser, confirm UI renders, toggles work, no console errors (other than expected favicon 404)
- [x] Settings modal: tabs switch, every slider updates its label, save/cancel/reset, escape closes
- [x] Per-conversation delete + Clear all history + history-list scroll
- [x] Stats panel: SERVER section populated from `/v1/models` + `/health`; MODEL/RUNTIME populated from `/api/show` after a successful generation; LAST RESPONSE updates per turn; SESSION counters update
- [x] Device badge: hidden until first response, then shows color-coded `CUDA · F16` / `CPU · F32` / etc. with informative tooltip
- [x] Sampling-summary chips reflect only active overrides
- [x] **Mock-driven multi-turn run** (5 turns in one conversation; mock SSE stream so test isn't blocked on local inference perf):
  - turn 1 greets and anchors memory ("name is Igor, color is teal")
  - turn 2 recalls both — request body contains 3 messages [user, assistant, user]
  - turn 3 streams a Python code block; rendered with `language-python` syntax highlight + Copy button
  - turn 4 emits `<think>...</think>` reasoning that becomes a collapsible disclosure
  - all four turns' stats footers populated (TTFT / tokens / tok/s / total)
- [x] Stop mid-stream: partial content kept, footer shows `aborted`, finally cleans up
- [x] Settings change applied to next turn — request body contains `temperature`, `top_k`, `min_p`, `repetition_penalty`, `seed`, `stop[]`
- [x] History persists across page reload — all turns + stats + reasoning + code blocks restored verbatim
- [x] Loading-state transitions verified with a real Phi-4-mini load: "Loading model Phi-4-mini-instruct · Ns" → cancel-load button appears → click cancel → server log shows `Model 'X' unloaded via stop request`, UI cleanly drops the placeholder
- [x] Error-notice differentiation verified with a real gemma-4 GGUF load failure (pre-existing model bug): rendered as a red-bordered notice, not as an Assistant turn

## Notes

- Pre-existing model bugs surfaced during testing (NOT introduced here, listed for visibility):
  - `ggml-org/gemma-4-26B-A4B-it-GGUF` — `tensor not found: model.language_model.layers.0.router.proj.weight`
  - `cyankiwi/gemma-4-26B-A4B-it-AWQ-4bit` — `Failed to parse config.json`
  - `unsloth/Phi-4-mini-instruct-GGUF` — missing `tokenizer.json` on HF (404)

- Follow-ups worth considering (out of scope here):
  - Populate the `quant_method` field from HF `quantization_config` for safetensors-AWQ / GPTQ models (right now we fall back to the model-id heuristic for those)
  - Stream usage info on the SSE response (`stream_options.include_usage`) so the per-message footer can be exact instead of "tokens ≈ chunks"


igorls:
Addressed all review feedback in `fbb7659`:

**Security**
- **Path traversal in `build.rs`** (Greptile P1, Gemini critical): URLs are now run through `sanitize_local_url` which rejects absolute paths and `..` components, plus `is_external` treats absolute paths (starting with `/`) as external so they're left in place rather than resolved against `ui/`. Query strings and fragments are stripped before path resolution.
- **XSS via `marked.parse`** (Greptile P2 / Gemini critical): added a `safeRenderer` that escapes raw HTML tokens (`<script>`, ``, etc.) and drops `javascript:` / `data:` / `vbscript:` URLs in links and images. Verified in-browser with a payload containing all three vectors — none execute, `window.__pwned*` stay undefined, `document.title` unchanged, but legitimate `**bold**` and `` `code` `` still render. Chosen over vendoring DOMPurify (~24 KiB) since the override is ~20 lines and covers the realistic LLM-output XSS surface.

**Correctness**
- **Attribute-order brittleness** (Greptile P2): rewrote the tag matcher to walk `<link>` / `<script>` opening tags, parse attributes individually, and require `rel="stylesheet"` (any token order) and `src=…` regardless of attribute order. `<link href="x" rel="stylesheet">` now bundles correctly.
- **Belt-and-suspenders assertion**: after bundling, the build fails if any `<link href>` / `<script src>` for a local URL survives — catches future parser misses at compile time instead of shipping a silently-broken UI.

**Lint**
- Ran `cargo fmt -p inferrs`; `cargo fmt --check -p inferrs` is now clean.

PR #249, opened 2026-08-14
Add `llmman sign` for cosign-compatible Sigstore signing

ericcurtin:
Attaches a cosign-compatible Sigstore signature to a model image as an OCI 1.1 referrer of its manifest:

  - Keyless (default): a Fulcio short-lived certificate for an OIDC identity, recorded in the public Rekor transparency log. The identity token comes from --identity-token, --identity-token-file, an ambient CI provider (GitHub Actions for now), or an interactive browser login, in that order.
  - Key-based (--key): ECDSA P-256 over an unencrypted PEM private key, packaged the same way `cosign sign --key` does. No network access beyond --remote's own registry round-trip.

Without --remote the signature is written into the local OCI store; with --remote the target manifest is resolved directly from the registry and the signature is pushed there under cosign's `sha256-<hex>.sig` fallback tag, alongside the image's own tag.

Keyless signing uses the `sigstore` crate (sigstore-rs); key-based signing is built by hand against `sigstore_protobuf_specs`' bundle types, since sigstore-rs only supports the keyless path today.

go-shim's llmman_inspect now returns the manifest's exact bytes instead of pretty-printing them, since `sign --remote` hashes that response to derive the digest it signs. `cmd::inspect` now does its own pretty-printing for display instead, so its output is unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
+dependencies = [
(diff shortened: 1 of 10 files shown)


ericcurtin:
Superseded by a rebased copy of this branch to resolve conflicts with main.

PR #254, opened 2026-08-19
fix(serve): stop a slow model cold-start from blocking every other model

doringeman:
ensure_model held state.0.manager's mutex across the whole cold-start path (subprocess spawn + wait_for_ready), not just around the map lookup/update. Since the fast path for an already-running model needs that same mutex, a request for model B queued behind model A's full load time whenever A happened to be loading concurrently — the opposite of the "each model gets its own subprocess" design.

Add a separate per-model load-lock registry (mirroring the existing MODEL_LOCKS pattern used for pull/push) that ensure_model holds across pull-if-missing/spawn/wait-until-ready instead of the manager mutex, which is now only ever taken briefly: once for the fast-path check, once more (after acquiring the load lock, in case another caller just finished) for the double-check, and once at the very end to publish the newly running model. Two concurrent requests for the same not-yet-loaded model still serialize on that model's load lock rather than each spawning a redundant backend process; unrelated models never contend on it at all.

Kept as a distinct lock registry (LOAD_LOCKS) rather than reusing MODEL_LOCKS, since ensure_model's load phase itself calls into pull_serialized, which takes a MODEL_LOCKS lock internally — sharing one map would re-enter the same non-reentrant tokio mutex from within the task already holding it.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -949,6 +949,18 @@ async fn ensure_model(state: &AppState, model_ref: &str) -> Result<u16, AppError
         let model_ref = canonical_ref(&state.0.store_path, model_ref);
         let model_ref = model_ref.as_str();
 
+        // Double-check again, under the (possibly different, now-resolved)
+        // post-pull name: another task may have raced us here under a
+        // different pre-pull alias for what turns out to be the same
+        // model (e.g. a bare shortname vs. its fully-qualified form, both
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

ericcurtin:
Check if bot comments are worth addressing, build must be green prior to merge, even if it's a flake, we need to fix the flake

ericcurtin (approved):

PR #255, opened 2026-08-20
docs: clarify which commands actually accept --store

doringeman:
The list (pull, push, transfer, rm, tag, inspect, serve) was already
stale — run, list, resolve, and launch resolve bare names too, just
via a different call path (client-side vs. the daemon's own /api/pull
handler). State the actual rule instead of a list that drifts out of
sync every time a new command starts resolving references.

"Default locations (override with --store <DIR>)" reads as if every
command takes the flag. Only the commands that touch the local store
directly (list, rm, tag, inspect, build, resolve, serve) do — pull,
push, run, launch, and ps always defer to whichever store the running
daemon was started with, and transfer/login/logout never touch a
local store at all. Verified against each command's actual clap args.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ llmman launch claude --model gemma4
 
 Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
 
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

ericcurtin:
Could potentially see us removing --store soon, might be more options than required

PR #257, opened 2026-08-21
test(e2e): give the opencode hang failure mode more retry headroom

doringeman:
CI run 32376290299 (PR #256, an unrelated 2-file llama_release.rs fix) hit launch_e2e.rs's already-documented endless-agent-loop failure mode (the model's sampling degenerating into a runaway decode loop that burns the whole 600s TIMEOUT, not a bad reply) on two independent platforms — Windows/docker and Linux/podman — in the same run, and exhausted all 3 retry attempts on both. That's for opencode specifically; claude and codex passed everywhere in the same run.

Bumped launch_e2e.rs's MAX_ATTEMPTS from 3 to 4. Since one test can now legitimately need up to 4 * 600s = 40 minutes in the worst case (vs. 30 before), raised the E2E job's timeout-minutes from 45 to 60 to keep the same headroom the original budget was reasoned out for — without this, a future run hitting the same failure mode across retries risks the outer job timeout hard-killing it mid-retry instead of letting launch_and_assert's own "giving up" path produce a clean panic and backtrace.

### DIFF
--- a/tests/launch_e2e.rs
+++ b/tests/launch_e2e.rs
@@ -549,7 +549,14 @@ fn run_launch(
 /// `llmman launch <integration> ...` invocation, against a fresh `HOME`
 /// each time, before giving up — see that function's own doc comment for
 /// why this exists at all.
-const MAX_ATTEMPTS: u32 = 3;
+///
+/// Was 3 until CI run 32376290299 (PR #256) hit the endless-agent-loop
+/// failure mode (see `launch_and_assert`'s doc comment) on two independent
+/// platforms (Windows/docker, Linux/podman) in the very same run — for
+/// `opencode` specifically, 3 attempts isn't reliably enough headroom.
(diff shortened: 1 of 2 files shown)


tests/launch_e2e.rs
```diff
@@ -549,12 +549,24 @@ fn run_launch(
 /// `llmman launch <integration> ...` invocation, against a fresh `HOME`
 /// each time, before giving up — see that function's own doc comment for
 /// why this exists at all.
-const MAX_ATTEMPTS: u32 = 3;
```
ericcurtin, line 552: Could we just make the change in this file simply?

const MAX_ATTEMPTS: u32 = 4;

would rather not complicate the code.

ericcurtin (approved):

tests/launch_e2e.rs
ericcurtin: Also a little curious why it needs so many retries and so many minutes

PR #262, opened 2026-08-24
fix(serve): default a missing tag before ensure_model's load lock

ericcurtin:
Fixes a race where concurrent requests for aliases of an unpulled
model (`gemma4` vs `gemma4:latest`) could take different load locks
and each spawn its own process.

`default_tag` appends `:latest` when missing, purely from the string
(mirrors ollama's `ParseName`). `ensure_model` now applies it before
acquiring the load lock, so both aliases key into the same lock.

Also: `cargo fmt` across the crate (it had drifted) plus a new
`cargo fmt --check` CI job so that doesn't happen again.

### release-note
```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,21 @@ env:
   LLAMA_CPP_RELEASE: b10293
 
 jobs:
+  fmt:
+    name: cargo fmt --check
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30 # stable
+        with:
+          components: rustfmt
+
+      - name: cargo fmt --check
+        run: cargo fmt --all -- --check
+
   build:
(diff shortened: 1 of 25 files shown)


ericcurtin:
Addressed the real gap: `default_tag` also normalized non-registry
sources (`ngc://`, `s3://`, `gs://`, `ms://`, absolute paths) by
appending a bogus `:latest`, since those can reach `ensure_model` via
`/api/chat`/`/api/generate`. Hardened it to no-op on those.

Left the other paths as-is:
- `handle_show`/`handle_delete`/the unload path already match aliases
  correctly — `OciStore::find`'s `ref_matches` applies the same
  tag-defaulting internally, and by definition a model must already be
  stored (canonical) to be shown/deleted/unloaded.
- `list`'s filter intentionally treats a tag-less filter as "match any
  tag" (see `matches_filter_matches_repo_without_tag`); forcing a
  default tag there would silently change that, not fix a bug.
- `handle_pull`/`handle_push`'s `MODEL_LOCKS` can still redundantly
  double-fetch for aliases racing a first pull — a pre-existing,
  documented tradeoff (see `MODEL_LOCKS`'s doc comment), not the
  spawn-duplication bug this PR fixes. Fixing it would mean changing
  what string crosses the FFI boundary into the Go shim's own
  `progressState` keying, a larger change I'd rather do separately
  with its own tests.

PR #263, opened 2026-08-24
run: match ollama run's Ctrl-C/Ctrl-D and thinking-block UX

ericcurtin:
Ports the rest of ollama's `cmd/cmd.go` streaming/display behavior into `llmman run` so it matches Ctrl-C, Ctrl-D, and the "thinking" highlighting as closely as possible.

## Ctrl-C

`chat_submit` is now async (`reqwest::Client`, not `::blocking`), racing the response stream against a single pinned `tokio::signal::ctrl_c()` per turn via `tokio::select!` — mirrors ollama's per-turn `context.WithCancel` + `signal.Notify`. In the interactive REPL, raw mode is released for the duration of each response and re-entered before the next prompt, like `readline.Instance.Readline()`. An interrupted turn's partial reply isn't added to history, matching `chat`/`generate` returning `nil` on cancel.

## Ctrl-D

Already correct from the existing readline port; re-verified through the new async plumbing.

## Thinking-block highlighting

Streamed output is word-wrapped at the terminal width, and `Thinking...`/`...done thinking.` render in ollama's own grey/bold ANSI (skipped when stdout isn't a terminal). New `--nowordwrap` flag mirrors ollama's own. A braille spinner now runs on stderr from submit until the first response chunk, matching `progress.NewSpinner`.

## Also

Interactive mode now also requires stdout to be a terminal, not just stdin, matching ollama's `RunHandler`.

## Testing

`cargo build`, `cargo fmt --all -- --check`, `cargo test --bin llmman` (81 passed) — 10 new unit tests for word-wrap and thinking-text formatting. Manually verified end-to-end against a mock `/api/chat` server: one-shot (spinner, colors, wrapping) and a pty-driven interactive session (Ctrl-C mid-stream cuts the response short and the session survives, Ctrl-D exits cleanly).

### DIFF
--- a/src/cmd/run.rs
+++ b/src/cmd/run.rs
@@ -1,8 +1,8 @@
 //! `llmman run` — interactive chat or one-shot prompt.
 //!
 //! Interactive mode uses a raw-mode readline ported directly from ollama's
-//! readline package (readline/readline.go, readline/term.go).  The key
-//! mechanism for paste detection mirrors ollama exactly:
+//! readline package (readline/readline.go, readline/term.go). Paste
+//! detection mirrors ollama exactly:
 //!
 //!   // ollama (Go)
 //!   if i.Terminal.reader.Buffered() > 0 { draining = true }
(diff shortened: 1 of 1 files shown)


ericcurtin:
Addressed the `ctrl_c()` reuse comment in 9c8569a — pinned a single listener for the whole turn instead of recreating it in each `select!`/loop iteration, which also happens to line up better with ollama's own `chat`/`generate` (one `signal.Notify` per turn, not per streamed chunk).

Not applying the word-wrap-when-redirected suggestion: I checked ollama's actual `cmd.go` (`RunHandler`) rather than just the `displayResponse` snippet — `opts.WordWrap = false` is set when stdin is piped (line ~806), but it's then unconditionally overwritten by `opts.WordWrap = !nowrap` a few lines later (~822), which runs regardless of that branch. So real `ollama run` currently *does* emit the same `\x1b[...D`/`\x1b[K` wrap escapes into a redirected/piped stdout unless `--nowordwrap` is passed — I confirmed this by tracing the control flow and by matching it against `llmman run model "prompt" > out.txt`'s current output, which reproduces the same escape sequences. Since the goal here is matching real `ollama run` behavior as closely as possible, "fixing" this would actually make `llmman` diverge from it. Happy to revisit if that turns out to be an ollama regression they fix upstream.

PR #270, opened 2026-08-25
fix(go-shim): make index.json writes atomic, reads lock-consistent

doringeman:
writeIndex overwrote index.json in place via a plain os.WriteFile. A process killed mid-write (or an unlucky page-cache flush ordering) left a truncated, permanently-corrupt index.json — breaking `llmman list`/`ps`/every future pull against that store until someone deleted or hand-repaired the file. Fixed by writing to a temp file and renaming it into place, the same atomic pattern writeBlob/ writeBlobStream already use for blob content in this same file — a reader can now only ever see the fully-old or fully-new content, never a partial mix.

Separately, updateIndex already took an exclusive advisory lock (lockIndex) around its own read-modify-write cycle, but pushToRegistry and cachedLayerName called readIndex directly, unlocked — so a concurrent updateIndex write could still be raced by a plain read that now (post atomic-write fix) can't observe corruption, but can still observe a stale-but-valid snapshot exactly as a write commits. Added readIndexLocked, taking the same lock briefly around a read-only readIndex call, and switched both call sites to it.

Checked ollama's own equivalent (manifest.WriteManifest in ../ollama/manifest/manifest.go): it has the identical gap — a plain os.Create with no rename, no locking — but its blast radius is smaller, since it writes one file per model rather than a single shared index every command depends on, so a torn write there only ever corrupts that one model's manifest. llmman's index.json is a single point of failure for the whole store, which is why this is worth the extra locking rigor here specifically. Ollama's own blob writer (manifest.NewLayer) already uses the same temp+rename pattern this fix applies to writeIndex, for the same reason.

Added shared_oci_test.go: a roundtrip/atomicity check for writeIndex, and a concurrent-updateIndex test (25 goroutines each adding a distinct manifest entry) verifying none of them get lost to the race this locking exists to prevent. Both pass under -race.

### DIFF
--- a/go-shim/backend_docker.go
+++ b/go-shim/backend_docker.go
@@ -428,7 +428,7 @@ func llmman_push(cLayoutDir, cRef *C.char) *C.char {
 // the last one).
 func pushToRegistry(ctx context.Context, layoutDir, ref string) (changed bool, err error) {
 	// Locate the manifest in the local index
-	idx, err := readIndex(layoutDir)
+	idx, err := readIndexLocked(layoutDir)
 	if err != nil {
 		return false, fmt.Errorf("read OCI index: %w", err)
 	}
--- a/go-shim/hf.go
+++ b/go-shim/hf.go
@@ -681,7 +681,7 @@ func parseHFRef(ref string) (host, owner, repo, tag string, err error) {
 // cachedLayerName returns the GGUF filename for ref if it is fully cached in
 // the local OCI store (manifest blob + all layer blobs present), or "" if not.
 func cachedLayerName(layoutDir, ref string) string {
-	idx, err := readIndex(layoutDir)
+	idx, err := readIndexLocked(layoutDir)
 	if err != nil {
 		return ""
 	}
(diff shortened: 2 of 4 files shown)


ericcurtin (approved):

ericcurtin:
Thanks for this, some feedback:

Ollama's implementation of this has one file per model, so there is less locking required, less contention, worth considering.
Docker/Moby's implementation of this has less of a dependency on locking files in general as a single dockerd owns the store. Can multiple independent llmman CLI processes touch this file at the same time? If yes maybe that's not right.
Podman's implementation has a single shared JSON index consumed by multiple independent processes like this, it is protected by flock plus atomic temp+rename.

Because we require a daemon for Ollama/OpenAI/Anthropic API capabilities. I would suggest further taking influence from Ollama and Docker/Moby in further PRs (moreso Ollama). Podman's daemonless design isn't as useful here, inference servers need a point of communication.

ericcurtin (approved):

PR #272, opened 2026-08-25
Replace --store flag with LLMMAN_MODELS and LLMMAN_TMPDIR env vars

ericcurtin:
## Summary

- Removes the `--store <DIR>` flag from `list`, `rm`, `tag`, `inspect`, `build`, `resolve`, `serve`.
- Adds `LLMMAN_MODELS` (mirrors Ollama's `OLLAMA_MODELS`), read directly by `default_store()`. Same default as before when unset/blank.
- Adds `LLMMAN_TMPDIR` (mirrors `OLLAMA_TMPDIR`), overriding where a `llama-server` release archive is staged before extraction.
- Hardens the download to never write through a pre-existing symlink at the staging path (relevant now that `LLMMAN_TMPDIR` can point at a shared directory and asset names are predictable).
- Updates README accordingly.

## Why

Every store-touching subcommand carried its own `store: Option<PathBuf>` field just to forward an override into `default_store` — inconsistent with every other `LLMMAN_*` setting in this codebase (`LLMMAN_CONTEXT_LENGTH`, `LLMMAN_FLASH_ATTENTION`, ...) and with Ollama's own env-var-only model. `pull`/`push` never had `--store` for the same reason.

## Testing

- `cargo build`, `cargo test --lib` (146 passed), `cargo fmt --check`, `cargo clippy` (no new warnings)
- Manually confirmed `--store` is gone from `--help` output

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,8 @@ Daemon-wide `llama-server` tuning, set before `llmman serve` starts:
 | `LLMMAN_CONTEXT_LENGTH` | Context size (`--ctx-size`) for every model this daemon loads. Defaults to a VRAM-tiered value when unset. |
 | `LLMMAN_FLASH_ATTENTION` | Flash Attention mode (`--flash-attn`): `on`, `off`, or `auto` (llama-server's own default). Also accepts `1`/`0`/`true`/`false`, matching Ollama's `OLLAMA_FLASH_ATTENTION`. |
 | `LLMMAN_KV_CACHE_TYPE` | KV-cache quantization (`--cache-type-k`/`--cache-type-v`), e.g. `f16` (default), `q8_0`, `q4_0` — trades output quality for memory at long context lengths, matching Ollama's `OLLAMA_KV_CACHE_TYPE`. |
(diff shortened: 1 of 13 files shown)


ericcurtin:
@coderabbitai review

ericcurtin:
@coderabbitai full review

ericcurtin:
@coderabbitai full review

ericcurtin:
Addressed the staging-cleanup-on-error and concurrent-download-collision comments. Skipping the `var_os` suggestion for `LLMMAN_MODELS`: every other `LLMMAN_*` var in this codebase (`LLMMAN_TMPDIR`, `LLMMAN_CONTEXT_LENGTH`, etc.) uses `env::var(...).ok()`, so keeping `LLMMAN_MODELS` consistent with that convention rather than special-casing it for a non-UTF-8-path edge case.

@coderabbitai full review

ericcurtin:
Declining the last suggestion (reusing the open file descriptor into `extract` instead of reopening by path): the remaining TOCTOU window is between our own exclusive-create download finishing and `extract`'s reopen, both in the same process, and closing it would mean reworking `extract_tar_gz`/`extract_zip`'s pre-existing path-based API beyond this PR's scope for a narrow, same-process race. Merging.

PR #276, opened 2026-08-25
Add cp/stop/show/gpu-discover commands, harmony + thinking parsers

ericcurtin:
## Summary

Ports several ollama-parity features that llmman was missing:

- **`llmman cp SOURCE DESTINATION`** — copies a local image under a new reference (same underlying operation as `tag`, just named/shaped like `ollama cp`).
- **`llmman stop MODEL`** — unloads a running model on demand, using the existing Ollama `keep_alive: 0` unload sentinel `handle_ollama_generate` already understands. Errors like `ollama stop` when the model isn't currently loaded.
- **`llmman show MODEL`** — architecture, parameter count, context length, quantization, license, and chat template for a locally stored model. Backed by a new minimal GGUF header reader (`src/gguf.rs`, metadata + tensor-info only, never tensor data) plus `config.json`/`tokenizer_config.json` reading for safetensors models. Supports `--license`/`--template`/`--parameters`/`--verbose`.
- **`llmman gpu-discover`** (hidden from `--help`, like ollama's own) — standalone entry point for the accelerator probe `llmman serve` already runs internally.
- **`src/thinking.rs`** — faithful port of ollama's `thinking/parser.go`: incremental, streaming-safe `<think>...</think>` extraction.
- **`src/harmony.rs`** — port of ollama's `harmony/harmonyparser.go`: the gpt-oss-style `<|start|>...<|channel|>...<|message|>...<|end|>` channel parser, plus a higher-level message handler splitting analysis/commentary/final channels into content/thinking/tool buckets.

The two parsers are wired into `cmd::serve`'s `/api/chat`/`/api/generate` streaming pipeline as a `RawContentExtractor` fallback: only engaged when a backend hands back raw, unparsed special tokens in `content` instead of already splitting them into a structured `reasoning_content`/`thinking` delta field (the existing, preferred path) — and once a backend has ever supplied structured thinking on a stream, the fallback is permanently disabled for the rest of it.

llmman already has `login`/`logout` (registry/HuggingFace credentials, same shape as `docker login`/`docker logout`), so no separate identity-keypair command is added here.

Also adds `fmt::human_count` and updates the README's command table.

## Review feedback addressed

- `RawContentExtractor` now buffers raw content until its mode is unambiguous (rather than guessing from a single, possibly one-byte chunk), correctly primes the harmony parser for a stream that starts mid-message at `<|channel|>`, and flushes any still-buffered prefix once the stream ends.
- `gguf.rs` bounds array-nesting recursion depth (avoids a stack overflow on a malformed file) and sums per-type element counts with saturating arithmetic (avoids an overflow panic).
- `show.rs` truncates long metadata on a UTF-8 char boundary (fixes a panic on non-ASCII metadata), no longer aborts the whole command when GGUF metadata is unreadable, and implements the `general.license*` GGUF fallback its own doc comment described.

## Testing

- `cargo build` / `cargo build --release` — clean, no new warnings.
- `cargo test --lib` — all green, including new regression tests for every fix above.
- `cargo fmt --check` — clean.
- Manual smoke tests: `gpu-discover`, `stop`/`cp`/`show` error paths, and a non-ASCII GGUF metadata file confirming the truncation fix no longer panics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,15 @@ Models are packaged as standard OCI artifacts and stored in any compatible regis
 | `resolve` | Pull (if needed) and print the local path of a model, as JSON — for other tools to consume |
 | `list`    | List locally stored models |
 | `ps`      | List models currently loaded |
+| `stop`    | Stop (unload) a running model |
 | `build`   | Package model files into a local OCI image |
 | `push`    | Push a local image to a registry |
 | `transfer` | Transfer an image directly from one location to another (e.g. HuggingFace to an OCI registry) |
+| `cp`      | Copy a local image to a new reference |
 | `rm`      | Remove a local image |
 | `tag`     | Create a new local tag pointing to an existing image |
 | `inspect` | Show the manifest of a local or remote image |
+| `show`    | Show a local model's architecture, parameters, license, and template |
 | `login`   | Log in to a container registry |
 | `logout`  | Log out from a container registry |
 
--- a/src/cmd/cp.rs
+++ b/src/cmd/cp.rs
@@ -1,0 +1,34 @@
+use clap::Args;
+
+use crate::storage::OciStore;
+
+#[derive(Args, Debug)]
+pub struct CpArgs {
+    /// Existing local reference to copy
+    #[arg(value_name = "SOURCE")]
+    pub source: String,
+
+    /// New reference to create, pointing at the same image
+    #[arg(value_name = "DESTINATION")]
+    pub destination: String,
+}
+
+/// `llmman cp SOURCE DESTINATION` — copy a model under a new name,
+/// mirroring `ollama cp` (`cmd/cmd.go`'s `CopyHandler`, `server/
+/// images.go`'s `CopyModel`): both are really just "point a second
+/// reference at the same content", exactly what `llmman tag` already does
+/// (see `storage::OciStore::tag`) — `cp` exists as its own command purely
+/// to match ollama's naming for anyone coming from there, not because the
+/// underlying operation differs at all.
+pub fn run(args: &CpArgs) -> anyhow::Result<()> {
+    let store_root = crate::default_store()?;
+    let store = OciStore::open(&store_root)?;
+
+    // resolve_ollama_api, not resolve: SOURCE must match how the model is
+    // actually stored — same reasoning as tag.rs/rm.rs.
+    let source = crate::shortnames::resolve_ollama_api(&args.source);
+    let desc = store.find(&source)?;
+    store.tag(desc, &args.destination)?;
+    println!("copied '{}' to '{}'", args.source, args.destination);
+    Ok(())
+}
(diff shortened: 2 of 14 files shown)


ericcurtin:
Addressed the actionable CodeRabbit feedback:

- **`RawContentExtractor` harmony detection** (major): now buffers raw content in `Undetermined` until the mode is unambiguous, instead of deciding from a single (possibly one-byte) chunk. Also primes the harmony parser with `add_implicit_start()` when a stream starts mid-message at `<|channel|>` rather than `<|start|>` (real llama-server completions routinely never re-emit the assistant preamble that was already part of the prompt). Added two regression tests: token-by-token streaming, and the `<|channel|>`-first case.
- **`show.rs` UTF-8 truncation panic** (major): fixed to truncate on a char boundary; verified with a manual non-ASCII smoke test (no longer panics).
- **`gguf.rs` unbounded array-nesting recursion** (major): bounded to 8 levels with a regression test constructing a 12-level-nested array that now errors cleanly instead of risking a stack overflow.
- **`show.rs` aborting on unreadable GGUF metadata** (minor): downgraded to a warning + `None`, rest of the output still prints.
- **`show.rs` `find_license`'s undocumented `general.license*` fallback** (minor): implemented as originally documented.
- **`identity.rs` missing-pubkey repair** (minor): `load_identity` now re-derives `id_ed25519.pub` from the private key if it's missing (e.g. an interrupted first `signin`), with a regression test. I did *not* add cross-process locking for the "two concurrent `signin` calls" race called out in that same comment — ollama's own `initializeKeypair` has the identical race (a bare `os.Stat` check, no lock either), and a full per-directory lock felt like more machinery than a rare, low-severity CLI race justifies here.
- **`fmt.rs` `human_count` boundary cases** (nitpick): added.
- **Cargo.toml `ssh-key`/`rand` versions** (nitpick): no change — that comment was advisory ("keep as-is, don't add RSA later"), not a request to change anything; `rsa`/`p256`/etc. are optional deps of `ssh-key` we don't enable (confirmed via `cargo tree -i rsa` — not part of the actual build graph despite appearing in `Cargo.lock`).

All 195 lib tests pass (up from 186 pre-PR), `cargo fmt --check` is clean, and release build succeeds.

ericcurtin:
Removed `signin`/`signout` from this PR per feedback — llmman already has `login`/`logout` (registry/HuggingFace credentials, same shape as `docker login`/`docker logout`), so a separate local identity-keypair command isn't needed alongside them. Also addressed the two follow-up review comments on the latest commit (saturating-add overflow in `gguf.rs`, and flushing `RawContentExtractor`'s buffered prefix at stream end) — see the updated PR description for the full list.

ericcurtin:
Addressed the latest CodeRabbit comment: `RawContentExtractor`'s transition to `Passthrough` (on the first chunk carrying backend-structured thinking) now flushes and prepends whatever `Undetermined` was still holding back for disambiguation, instead of silently dropping it — the one path the earlier `flush`-on-`done` fix didn't already cover. Added a regression test. All 196 lib tests pass, `cargo fmt --check` clean, release build succeeds.

PR #281, opened 2026-08-25
fix(vllm-plugin): patch vLLM 0.27's speculators check for oci:// refs

ericcurtin:
## Problem

`main`'s CI has been failing on every push since #274 (the PR that added `vllm-plugin`'s real e2e coverage). Every single CI run since then fails identically, on every OS/backend in the matrix, at the same step:

```
pytest -m e2e (vllm-plugin)
...
huggingface_hub.errors.HFValidationError: Repo id must be in the form 'repo_name' or
'namespace/repo_name': 'oci://qwen3.5:0.8b-safetensors'. Use `repo_type` argument if needed.
```

This isn't a flake — it reproduces 100% of the time against the pinned `VLLM_VERSION: 0.27.1`.

## Root cause

vLLM 0.27 added a second HuggingFace touchpoint that runs *before* `ModelConfig` is ever constructed:

```python
# vllm/engine/arg_utils.py, EngineArgs.create_engine_config
if not is_cloud_storage(self.model):
    (self.model, self.tokenizer, self.speculative_config) = (
        maybe_override_with_speculators(model=self.model, tokenizer=self.tokenizer, ...)
    )
model_config = self.create_model_config()  # <- our existing patch only fires in here
```

`maybe_override_with_speculators` calls `PretrainedConfig.get_config_dict(model)` directly on the raw string. It's guarded by `is_cloud_storage()`, which vLLM only wires up for `s3://`/`gs://`/`az://` — not `oci://` — so it runs unconditionally for `oci://` refs and crashes trying to treat `oci://qwen3.5:0.8b-safetensors` as an HF repo id, well before our existing `ModelConfig.maybe_pull_model_tokenizer_for_runai` patch ever gets a chance to resolve it.

## Fix

Same technique the plugin already uses for `maybe_pull_model_tokenizer_for_runai`: wrap `vllm.engine.arg_utils.maybe_override_with_speculators` at `install()` time, and short-circuit it (return the refs unchanged) whenever `model`/`tokenizer` is an `oci://` reference, delegating to the original for everything else. The real resolution still happens later, unchanged, inside `maybe_pull_model_tokenizer_for_runai`.

## Verification

- Added unit tests for the new `_make_patched_speculators` (mirrors the existing `_make_patched` tests) and an integration test confirming `install()` replaces the real `vllm.engine.arg_utils.maybe_override_with_speculators`.
- `pytest tests/` (unit, not `-m e2e`): 24 passed.
- Manually reproduced the exact crash against a real `vllm==0.27.1` install before this change, confirmed it's gone after.
- Ran the actual failing e2e test end-to-end locally: `vllm serve oci://qwen3.5:0.8b-safetensors ...` comes up healthy and a real `/v1/chat/completions` request against it replies correctly.

### DIFF
--- a/vllm-plugin/src/vllm_llmman/_patch.py
+++ b/vllm-plugin/src/vllm_llmman/_patch.py
@@ -14,13 +14,28 @@ runtime. See vLLM's `vllm/config/model.py` (`__post_init__`, line ~565)
 and `vllm/transformers_utils/runai_utils.py` for the method being
 wrapped.
 
-`_make_patched` is a pure function (no vLLM import) so it's unit
-testable against a stub object; `install()` is the only piece that
-actually touches `vllm.config.model.ModelConfig`.
+vLLM 0.27 added a second, *earlier* HuggingFace touchpoint:
+`EngineArgs.create_engine_config` calls `maybe_override_with_speculators`
+directly on the raw `model`/`tokenizer` strings, before `ModelConfig` (and
+therefore `maybe_pull_model_tokenizer_for_runai`) is ever constructed. It
+guards this with `is_cloud_storage()`, which it uses to skip `s3://`/
+`gs://`/`az://` — but that helper doesn't know about `oci://`, so an
+unpatched vLLM 0.27+ calls `PretrainedConfig.get_config_dict("oci://...")`
+and blows up before our other patch gets a chance to run. See vLLM's
+`vllm/engine/arg_utils.py` (`create_engine_config`, line ~1917) and
+`vllm/transformers_utils/config.py` (`maybe_override_with_speculators`).
+We wrap that function for `oci://` `model` refs, short-circuiting
+speculators config loading so the real resolution still happens later,
+inside `maybe_pull_model_tokenizer_for_runai`.
+
+Both `_make_patched` and `_make_patched_speculators` are pure functions
+(no vLLM import) so they're unit testable against stubs/fakes; `install()`
+is the only piece that actually touches vLLM's modules.
 """
 
 from __future__ import annotations
 
+import inspect
 import logging
 from typing import Any, Callable, Protocol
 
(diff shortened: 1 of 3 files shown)


ericcurtin:
Addressed both CodeRabbit findings in the amended commit:

1. `_make_patched_speculators` now only short-circuits on `is_oci_ref(model)` — confirmed against vLLM's actual `maybe_override_with_speculators` that it only ever reads speculators config from `model`, so an `oci://` `tokenizer` alongside a real HF `model` isn't this crash and must still go through real speculators detection. Updated the corresponding test to assert delegation instead of short-circuiting in that case.
2. `test_install_integration.py`'s fixture now treats `maybe_override_with_speculators` as optional (`getattr(..., None)`) and only restores/deletes it when present; the speculators-specific test now skips (rather than erroring) if a future/older vLLM doesn't have that hook.

Re-ran the full non-e2e suite (24 passed) after both fixes.

ericcurtin:
Addressed the latest CodeRabbit finding, with one correction:

The report that vLLM calls `maybe_override_with_speculators` positionally with a third `trust_remote_code` argument doesn't match reality — I checked, and `create_engine_config` (the only call site in vLLM 0.27.1) passes every argument by keyword. That said, the underlying concern is legitimate: `patched`'s own hand-rolled `(model, tokenizer=None, **kwargs)` signature didn't actually match `original`'s real signature, so it would've broken if anything ever did call it positionally (or with `trust_remote_code` passed positionally, third-party callers, a future vLLM refactor, etc).

Fixed by binding against `original`'s own `inspect.signature()` instead of hand-rolling `patched`'s parameter list, so any calling convention `original` itself accepts (positional or keyword) keeps working transparently through the wrapper. Added a regression test calling `patched(model, tokenizer, trust_remote_code)` positionally, and updated the other speculators tests' fake `original` to match the real function's actual signature (previously a bare `**kwargs` catch-all, which would've hidden this exact class of bug).

25 passed locally after this change.

ericcurtin:
Fixed the module docstring: it said this patches for `oci://` "model/tokenizer refs", which was stale from before the earlier fix restricted the check to `model` only. Now says `oci://` `model` refs, matching the actual (correct) behavior.

25 passed locally after this change.

PR #282, opened 2026-08-25
Revert: remove the hf-xet __xet-fetch subcommand

ericcurtin:
Reverts #280.

That PR was explicitly scoped as a foundation only: `llmman __xet-fetch` worked end-to-end but was never wired into `transfer`/`pull` on the Go side, so it had zero effect on any real code path — it just added a dependency tree (`hf-xet`, `xet-client`, `xet-data`, `xet-runtime`) and a hidden subcommand nothing called.

This removes it so things go back to exactly how they worked before #280: plain-HTTP downloads (with the `Range` header fix from #278 and `Retry-After` handling from #279, both untouched by this revert) for everything, no `hf-xet` dependency, no hidden subcommand.

Clean revert of `3054d42`, no conflicts. `cargo build`/`cargo test --lib` pass (196 tests, down from 201 — the 5 removed were `xet_fetch`'s own).


ericcurtin:
Taking a different approach: removing the discoverable __xet-fetch subcommand but keeping (and actually wiring in) the underlying hf-xet integration, so it's used transparently by transfer/pull rather than not at all. Superseding PR to follow.

PR #284, opened 2026-08-25
feat: port HuggingFace pull/transfer to Rust, wire in real Xet support

ericcurtin:
## Why

HF pull/transfer logic lived entirely in the Go shim, using plain HTTP GETs even for Xet-CAS-backed files — a CDN-bridge path `huggingface_hub` itself refuses past ~50GB, with no chunk-level dedup or resume. #280 added `hf-xet` downloads via a hidden `__xet-fetch` subcommand the shim would self-exec, but that boundary was wrong: every consumer of an HF file's bytes needs to decide per-file whether to stream via Xet or HTTP, and self-exec per file was slow and a bad fit for transfer's live registry push (needs the blob's digest known before the first byte is sent, no disk touched).

## What

Ports pull/transfer to Rust instead, calling `hf-xet` in-process, keeping only actual containerd/podman registry-protocol calls in Go. Cloud SDK sources (ModelScope, NGC, S3, GCS) stay Go-only — no test credentials to verify a port this round, and they don't touch Xet.

New `src/hf/` module tree: `mod.rs` (auth, host/ref classification), `client.rs` (retry/backoff, the Rust side of #279), `api.rs` (HF file listing, GGUF/mmproj/license selection), `download.rs` (`head_metadata`, `fetch_once`), `oci.rs` (local OCI layout, CNCF manifest building), `progress.rs` (native progress tracking), `pull.rs`/`transfer.rs` (orchestration).

For transfer's registry push to stream straight through with no local disk, Go hands Rust a live pipe: Go creates it (proven cross-platform in its own runtime) and passes only the write end's raw fd/HANDLE across the existing Rust-calls-Go FFI direction, which Rust wraps via `File::from_raw_fd`/`from_raw_handle`. New `go-shim/push_stream.go` (+ `_unix`/`_windows`, `!podman`-tagged) scopes one resolver/pusher per whole transfer, not per blob — per-blob caused real "broken pipe" errors against a slow-to-negotiate registry.

## Bugs fixed

Two found only via a live end-to-end transfer:
- `head_metadata` used a redirect-following client, so it read the *final* CDN response's headers instead of the huggingface.co hop carrying `X-Linked-Etag`/`X-Xet-Hash`, misreading the Xet hash as the digest. Fixed with a dedicated no-redirect client. Regression-tested.
- `FileMetadata::digest` comes back bare hex; the OCI `Descriptor` needs the `"sha256:"` prefix added, or the registry rejects it as `DIGEST_INVALID`.

Also fixed from review feedback on this PR:
- Two temp-file races (`pull.rs`, `oci.rs::write_blob`) where concurrent writes with colliding filenames/digests could corrupt each other — added a per-call unique counter.
- `select_gguf`'s shard splitter could silently return an incomplete GGUF split; now errors instead.
- A user-requested GGUF tag/quant that doesn't exist silently fell back to safetensors; now only falls back when the repo has no GGUF files at all.
- `head_metadata` returned an empty, "small file"-shaped result after exhausting its redirect-hop budget — risking buffering a huge file in memory on `transfer`'s path; now errors instead. Wrapped both callers' `head_metadata` calls in the existing retry helper too.
- `retry_delay` could underflow/overflow its shift for attempt 0 or ≥32; clamped.
- The new pipe wasn't marked `FD_CLOEXEC`, so a future fork/exec during a stream could leak the write end and hang `push_stream_wait` forever; fixed with `syscall.CloseOnExec`.
- `login()` wrote the token file world-readable, then chmod'd it 0600 after; now created with mode 0600 directly.
- `is_oci_registry` only probed HTTPS and only recognized the standard header, so a plain-HTTP or challenge-only registry given as a pull/transfer *source* misclassified as HuggingFace; now also tries HTTP and accepts a `401`+`WWW-Authenticate` challenge.
- `stream_ffi_progress` preferred an empty Go snapshot over a real native one; fixed the fallback order.

Also fixed `shortnames::has_host` to recognize `host:port`, found while testing against a local registry.
## Testing (updated)

Verified end-to-end against a local `registry:2` after every fix, including two rounds of review fixes: `llmman transfer hf.co/hf-internal-testing/tiny-random-gpt2 localhost:5959/...` and the `hf://owner/repo` scheme form, both fresh and as a same-content repeat transfer (exercises the registry's "already exists" path). Every pushed blob's digest checked via `llmman resolve` and a direct registry GET + `shasum`. `cargo test --lib`: 250 passed. `go build`/`vet`/`test`: clean. Both `docker` (default) and `podman` feature builds compile. CI green on all platforms (including both Windows targets).

## Deferred

Cloud SDK sources stay Go-only. Go's now-HF-dead pull code (`hf.go` etc.) stays in place — still shared by those sources' helpers. A few lower-value review nits (cross-model blob dedup before download, blocking-I/O isolation for `stream_to_writer`'s pipe writes, `hf-xet` resource reuse across retries, network tests not gated behind `#[ignore]`) left as follow-ups — see PR comments for the full triage.


ericcurtin:
Pushed fixes for the review feedback above. Summary:

**Fixed** (all re-verified with a live end-to-end transfer against a local registry, plus `cargo test --lib`/`go test ./...`):
- Temp-file races in `pull.rs`/`oci.rs::write_blob` (concurrent writes could corrupt each other)
- `select_gguf` silently returning an incomplete shard split
- A requested GGUF tag that doesn't exist silently falling back to safetensors
- `head_metadata` returning an empty "small file" result after exhausting its hop budget (risked buffering a huge file in memory) — now errors, and both callers retry it
- `retry_delay`'s shift under/overflow for attempt 0 or ≥32
- Missing `FD_CLOEXEC` on the new pipe (`push_stream_unix.go`)
- Token file written world-readable before chmod
- `is_oci_registry` missing plain-HTTP/challenge-only registries (misclassified a local registry given as a pull/transfer source)
- `stream_ffi_progress` preferring an empty Go snapshot over a real native one

**Deferred as follow-ups** (lower value relative to risk of a rushed fix here):
- Cross-model blob dedup (`oci::blob_exists` check before downloading) — a real optimization, not a correctness bug; the existing per-reference `report_cached` already skips a full repeat pull.
- Blocking file I/O isolation via `spawn_blocking` for the CLI's direct `resolve` path — flagged by the bot itself as a heavy lift; the daemon path already has this via its own `spawn_blocking` boundary, and a single CLI invocation doesn't contend with anything else.
- Reusing `hf-xet`'s `XetSession`/`XetDownloadStreamGroup` across retries/files — perf optimization, not correctness.
- `via_temp_pull` (podman) always reporting `changed=true` — documented the limitation instead of restructuring `ffi::push`'s return contract, since it's cosmetic output only.
- `classify()`'s handling of bare, unqualified multi-segment refs — the actual root cause is `shortnames::resolve`'s pre-existing, unconditional `hf.co/` defaulting for any hostless reference regardless of segment count, which predates this PR and is out of scope here.

ericcurtin:
Pushed a second round of fixes from the latest review. The most important one: **a repeat transfer of an already-pushed model always failed** — `pushLazy` returns as soon as the registry says it already has a blob, without ever reading the pipe Rust writes into, so Go closed the read end out from under an in-progress (or about-to-start) write, guaranteeing a broken pipe on every retry. I reproduced this against a local registry before fixing it (push twice, same content) and confirmed it's now fixed the same way.

Also fixed:
- `classify()` returning a hostless ref for the documented `hf://owner/repo` scheme form, which `parse_hf_ref` can't parse — every scheme-form pull/transfer without an explicit host was broken.
- GGUF shard matching being case-sensitive against a lowercased extension check (a `.GGUF` split could silently parse as unsplit).
- Shard grouping ignoring the parent directory (same-named shards under different quant subdirectories could merge or miscount); duplicate indices are now rejected too.
- A zero-size `Descriptor` being possible on `transfer`'s streaming path if both `head_metadata` and the file listing failed to report a size.
- `pull.rs` never checking its own computed digest against the server-declared one from `head_metadata`.
- `PushStream` missing `#[must_use]`, and the new pipe's creation missing `syscall.ForkLock.RLock` (matching what `os.Pipe` itself does on unix).
- `head_metadata` building a new HTTP client per call instead of a shared one.

Re-verified end-to-end after all of the above (including the previously-broken `hf://` scheme form and a same-content repeat transfer). `cargo test --lib`: 250 passed. `go build`/`vet`/`test`: clean.

ericcurtin:
Latest review's two remaining comments (`is_hf_host` not stripping a port, `login()` chmod'ing an existing shared `HF_TOKEN_PATH` parent directory) both concern code that predates this PR unchanged — it's carried over verbatim from the already-merged #280's `src/hf.rs` into this PR's `src/hf/mod.rs` split, not something introduced or touched by the porting work itself. Valid findings, but out of scope here; worth a small follow-up PR rather than scope-creeping this one further.

CI is green on every platform (including both Windows targets, build + unit tests), all prior review rounds addressed and re-verified end-to-end. Merging.

PR #291, opened 2026-08-26
fix(daemon): fail fast when llmman serve dies at startup

### LINKED ISSUE #289: Silent 60s hang when the serve daemon dies at startup
## Summary

If the background `llmman serve` daemon exits during startup, the client polls a dead port for 60s, prints nothing, then fails with only "llmman serve did not start within 60s". The real error sits in serve.log, which the message never mentions.

## Reproduction

With no `llama-server` on PATH and a failing auto-download:

    llmman launch claude --model gemma4

60 seconds of silence, then the generic timeout. Meanwhile `~/.local/share/llmman/serve.log` had the cause the whole time:

    Error: no llama-server on PATH and automatic download failed: no vulkan
    llama.cpp release asset found in v0.3.0 (...)

The daemon had exited within a second of spawning.

## Root cause

[`ensure_server`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/daemon.rs#L200-L275) spawns the daemon detached, then [discards the `Child` and polls `server_alive()` 120 times at 500ms](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/daemon.rs#L266-L274). It never checks whether the child exited, so any daemon startup failure burns the full 60s and surfaces as a generic timeout. All of [`launch`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/cmd/launch.rs#L54), [`run`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/cmd/run.rs#L123), [`pull`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/cmd/pull.rs#L19), and [`push`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/cmd/push.rs#L19) share this helper.

maxgio92:
Closes #289.

[`ensure_server`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/daemon.rs#L200-L275) spawned the daemon detached, [discarded the `Child`, and polled `server_alive()` 120 times at 500ms](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/daemon.rs#L266-L274). A daemon that exited during startup (for example, a failed llama-server auto-download) was never detected: the client polled a dead port for the full 60s, printed nothing, and reported only "llmman serve did not start within 60s" while the real error sat unmentioned in serve.log.

Changes, all in `src/daemon.rs`:

- Keep the `Child` from `spawn()` and call `try_wait()` inside the poll loop. If the daemon exited, bail immediately with its exit status and the last five non-empty lines of serve.log. The daemon is in its own process group but is still our child, so `try_wait()` works on all platforms.
- Run one final `try_wait()` before the timeout bail, so an exit landing after the last poll still reports the status instead of the generic timeout.
- New `log_tail` helper formats the serve.log excerpt: empty when there is no log path, just the path when the file is unreadable or empty. The timeout message now uses it too, so users know where to look.

Drive-by: `cargo clippy --all-targets` was failing on main ([`non_octal_unix_permissions` in a test](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/storage/oci.rs#L977)); changed `from_mode(0)` to `from_mode(0o0)`.

## Testing

- 3 new unit tests for `log_tail`; full lib suite passes (255 tests), clippy clean.
- Verified end-to-end on a machine where the daemon dies at startup: `llmman pull gemma4` went from 60s of silence to failing in ~0.5s with:

      Error: llmman serve exited during startup (exit status: 1); last lines of /home/.../serve.log:
      Error: no llama-server on PATH and automatic download failed: no vulkan llama.cpp release asset found in v0.3.0 (...)

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -421,30 +421,102 @@ pub fn ensure_server(preload_model: &str) -> anyhow::Result<()> {
     for _ in 0..120 {
         std::thread::sleep(Duration::from_millis(500));
         if server_alive() {
+            // The connect proves the daemon was alive an instant ago, not
+            // that it survived: if it died right after accepting, report
+            // its real exit status instead of an Ok the caller's next
+            // request would contradict. A death after this check is
+            // inherently unobservable from here.
+            bail_if_exited(&mut child, log_path.as_deref())?;
             return Ok(());
         }
         // The daemon is in its own process group but still our child, so
         // try_wait catches an immediate startup failure (e.g. llama-server
         // auto-download failing) instead of polling a dead port for 60s.
-        if let Some(status) = child.try_wait().context("wait on llmman serve")? {
-            anyhow::bail!(
-                "llmman serve exited during startup ({status}){}",
-                log_tail(log_path.as_deref())
-            );
-        }
+        bail_if_exited(&mut child, log_path.as_deref())?;
     }
     // The daemon may have exited right after the final poll above: check
     // once more so that narrow window still reports the exit status
     // instead of the generic timeout.
-    if let Some(status) = child.try_wait().context("wait on llmman serve")? {
-        anyhow::bail!(
-            "llmman serve exited during startup ({status}){}",
-            log_tail(log_path.as_deref())
-        );
+    bail_if_exited(&mut child, log_path.as_deref())?;
+    // The last in-loop probe is up to 500ms stale by now; a daemon that
+    // bound in that window is a healthy start, not a timeout.
+    if server_alive() {
+        return Ok(());
     }
+    // Timed out with the daemon alive but not listening: stop it before
+    // reporting failure, rather than leave a half-started daemon running
+    // detached after the user was told the start failed. TERM the group
+    // first, then escalate to a group KILL after a short grace (like
+    // stop_stale_daemon, so a slow-to-exit child is actually reaped too),
+    // then wait so no zombie outlives this call. The message must say the
+    // daemon was stopped: a slow first run (llama-server's one-time
+    // download can exceed this budget) would otherwise converge on a
+    // plain retry, and now it will not.
+    signal_group(child.id(), false);
+    std::thread::sleep(Duration::from_millis(500));
+    signal_group(child.id(), true);
+    let _ = child.wait();
     anyhow::bail!(
-        "llmman serve did not start within 60s{}",
+        "llmman serve did not start within 60s and was stopped; run 'llmman serve' in the \
+         foreground to watch what startup is doing (e.g. a first-time llama-server download){}",
         log_tail(log_path.as_deref())
+    )
+}
+
+/// Best-effort signal to the daemon's whole process group (the daemon is
+/// its own group leader, see detach), with no plain-pid fallback: unlike
+/// kill_daemon's stale-daemon path, callers here may hold an
+/// already-reaped pid, and a single-pid signal to a freed pid is the one
+/// form that could hit an innocent reused process. Failures are ignored;
+/// an already-empty group is the common case.
+#[cfg(unix)]
+fn signal_group(pid: u32, force: bool) {
+    let signal = if force { "-KILL" } else { "-TERM" };
+    let _ = Command::new("kill")
+        .args([signal, &format!("-{pid}")])
+        .status();
+}
+
+/// Windows equivalent: taskkill's /T already targets the whole process
+/// tree, and fails harmlessly on an already-dead pid.
+#[cfg(windows)]
+fn signal_group(pid: u32, force: bool) {
+    let mut args = vec!["/PID".to_string(), pid.to_string(), "/T".to_string()];
+    if force {
+        args.push("/F".to_string());
+    }
+    let _ = Command::new("taskkill").args(&args).status();
+}
+
+/// Bails with the daemon's exit status (and its log tail) if the freshly
+/// spawned `llmman serve` child has already exited; returns Ok(()) while
+/// it's still running. On the exited branch this also best-effort kills
+/// the daemon's whole process group first, so a llama-server child the
+/// daemon spawned before dying is not left holding a loaded model.
+fn bail_if_exited(
+    child: &mut std::process::Child,
+    log_path: Option<&std::path::Path>,
+) -> anyhow::Result<()> {
+    let pid = child.id();
+    let Some(status) = child.try_wait().context("wait on llmman serve")? else {
+        return Ok(());
+    };
+    // The daemon may have spawned a llama-server before dying, and that
+    // child shares the daemon's process group (see detach), so signal the
+    // group rather than orphan a loaded model. The leader itself is dead
+    // and reaped by now; only the group form of the signal is meaningful
+    // (see signal_group on why there is deliberately no single-pid
+    // fallback here).
+    signal_group(pid, false);
+    // Reading the log now needs no explicit sync: a daemon that fails via
+    // its error-exit path reports on stderr, which Rust never buffers,
+    // and that fd is redirected straight to the log file, so the reason
+    // is visible to any reader before the exit is observable. A daemon
+    // killed by a signal writes nothing; log_tail's "(see path)" fallback
+    // covers that.
+    anyhow::bail!(
+        "llmman serve exited during startup ({status}){}",
+        log_tail(log_path)
     )
 }
 
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -472,8 +472,12 @@ pub fn ensure_server(preload_model: &str) -> anyhow::Result<()> {
 #[cfg(unix)]
 fn signal_group(pid: u32, force: bool) {
     let signal = if force { "-KILL" } else { "-TERM" };
+    // Stdio nulled: an already-empty group makes kill print "No such
+    // process", which would land in front of the real startup error.
     let _ = Command::new("kill")
         .args([signal, &format!("-{pid}")])
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
         .status();
 }
 
@@ -485,7 +489,13 @@ fn signal_group(pid: u32, force: bool) {
     if force {
         args.push("/F".to_string());
     }
-    let _ = Command::new("taskkill").args(&args).status();
+    // Stdio nulled like the Unix branch: taskkill reports a missing PID
+    // loudly, which would land in front of the real startup error.
+    let _ = Command::new("taskkill")
+        .args(&args)
+        .stdout(Stdio::null())
+        .stderr(Stdio::null())
+        .status();
 }
 
 /// Bails with the daemon's exit status (and its log tail) if the freshly
@@ -501,6 +511,13 @@ fn bail_if_exited(
     let Some(status) = child.try_wait().context("wait on llmman serve")? else {
         return Ok(());
     };
+    // An exited child is not always a failed startup: two concurrent
+    // clients can both spawn a daemon, and the loser exits with "address
+    // in use" while the winner serves. If something is listening now, the
+    // start succeeded, whoever's daemon it is.
+    if server_alive() {
+        return Ok(());
+    }
     // The daemon may have spawned a llama-server before dying, and that
     // child shares the daemon's process group (see detach), so signal the
     // group rather than orphan a loaded model. The leader itself is dead
@@ -1089,7 +1106,9 @@ mod tests {
         let tail = log_tail(Some(&path));
         std::fs::remove_file(&path).unwrap();
         assert!(tail.contains("last lines of"), "got: {tail}");
-        assert!(!tail.contains("one"), "got: {tail}");
-        assert!(tail.contains("two\nthree\nfour\nfive\nsix"), "got: {tail}");
+        // Only the part after the header line is the log excerpt; the
+        // header contains the temp path, which may itself contain "one".
+        let excerpt = tail.split_once('\n').map(|(_, rest)| rest).unwrap_or("");
+        assert_eq!(excerpt, "two\nthree\nfour\nfive\nsix", "got: {tail}");
     }
 }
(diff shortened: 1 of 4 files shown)


ericcurtin:
Thanks for picking this up. Need to fix formatting to pass the build. There are also some conflicts to fix. Also, pick up any review comments from coderabbit and copilot worth picking up. A few review notes:

1. **Duplicated bail block** (`daemon.rs:276-281` vs `286-291`): the `try_wait` + `anyhow::bail!` pair is copy-pasted for the in-loop check and the post-loop check. Worth factoring into a small helper to avoid drift.

2. **Orphaned `llama-server` on early exit**: if the daemon already spawned its own `llama-server` child before dying, the fast-fail path here only reaps the `llmman serve` process — the process group isn't killed the way `kill_daemon`/`stop_stale_daemon` does for a stale daemon. Could leak a loaded-model subprocess on every failed startup past that point. Should this path also clean up the process group before bailing?

3. **Log flush race**: `log_tail` reads `serve.log` right after `try_wait()` confirms exit. This only shows the crash reason if the daemon's log writer is unbuffered/flushed on exit. It's worth confirming that's actually the case, since it's the assumption the whole feature relies on.

4. **Test coverage**: the 3 new tests only cover `log_tail` in isolation, not `ensure_server`'s new `try_wait` branch itself. Given `tests/launch_e2e.rs` already spawns `llmman` end-to-end, a test with a stub "serve" that exits immediately would guard the actual behavior change (fast-fail + message), not just string formatting.

maxgio92:
Thanks @ericcurtin 

I'm going to address 1, 2, and 4. But also points #3, because I do think that it should reap the process group, you're right I believe.
I think that in this case this fixes operate on, the logs are flushed when buffered, but definitely worth double checking edge cases.

ericcurtin:
> Thanks @ericcurtin
> 
> I'm going to address 1, 2, and 4. But also points #3, because I do think that it should reap the process group, you're right I believe. I think that in this case this fixes operate on, the logs are flushed when buffered, but definitely worth double checking edge cases.

Makes sense to me

src/daemon.rs
```diff
@@ -416,15 +416,54 @@ pub fn ensure_server(preload_model: &str) -> anyhow::Result<()> {
...
+    if let Some(status) = child.try_wait().context("wait on llmman serve")? {
+        anyhow::bail!(
+            "llmman serve exited during startup ({status}){}",
+            log_tail(log_path.as_deref())
+        );
+    }
+    anyhow::bail!(
+        "llmman serve did not start within 60s{}",
+        log_tail(log_path.as_deref())
+    )
```
ericcurtin, line 474: On this timeout path (and similarly on the two `exited during startup` bails above), we return an error but never touch `child` when it's still alive. For the timeout case specifically, the daemon process we just spawned is left running, detached, in the background — the user sees "did not start within 60s" but the process (and, once it gets that far, any `llama-server` it spawns) keeps running and holding the port/GPU/etc.

Since this PR is already threading the `Child` through, could we kill it (and its process group on Unix, matching `detach`'s own note about the daemon being in its own group) before bailing here? Something like a small helper called from both the timeout branch and the immediate-exit branches (the latter is arguably a no-op if the process already exited, but doesn't hurt to reap it explicitly rather than relying on `try_wait` having already reaped it).

This isn't a regression introduced by this PR — the previous code discarded the `Child` right after `spawn()` and leaked identically on timeout — but it seems worth fixing now while we're already holding the handle, rather than carrying the leak forward into the new fail-fast path.

src/daemon.rs
```diff
@@ -416,15 +416,54 @@ pub fn ensure_server(preload_model: &str) -> anyhow::Result<()> {
         }
     }
     detach(&mut cmd);
-    cmd.spawn().context("spawn llmman serve")?;
+    let mut child = cmd.spawn().context("spawn llmman serve")?;
 
     for _ in 0..120 {
         std::thread::sleep(Duration::from_millis(500));
         if server_alive() {
```
ericcurtin, line 423: Minor/non-blocking: `server_alive()` is checked before `child.try_wait()` each iteration, so in the (very narrow) window where the daemon's process exits in the same tick it briefly had the port open, we'd report success and return `Ok(())` without ever having validated the child is actually still running. Given `server_alive()` is a real TCP connect (not just a stale socket check), I think this is fine in practice — just flagging in case it was considered and intentionally left as a non-issue rather than something to change.

ericcurtin (commented):
Reviewed the diff directly (not just the bot output). This is a solid, well-scoped fix: keeping the `Child` handle and polling `try_wait()` is exactly right, and I like that the final post-loop check closes the race where the daemon dies right after the last poll. The `log_tail` helper and its three unit tests read cleanly, and the `serve.log` it reads is freshly rotated/truncated for this daemon instance, so reading the whole file in one shot is fine — no need for a bounded/streaming read there despite what one of the automated reviews suggested.

One real gap worth addressing before/after merge (see inline comment): on the 60s-timeout path the still-running daemon process is never terminated. Since this PR is the one introducing the retained `Child` handle, it's a natural place to also clean it up. Not a regression — the old code discarded the handle immediately and leaked the same way — but now that we hold the handle, leaving it running seems like an oversight rather than a choice.

CI is green across all targets and clippy is clean (including the drive-by `0` -> `0o0` fix, which is a no-op behaviorally and just satisfies `non_octal_unix_permissions`).

src/daemon.rs
maxgio92: Good catch! Going to address it. Thank you

src/daemon.rs
maxgio92: I've found an edge case where it could happen: bind-then-die-during-init used to yield Ok and a bare connection refused on the next request; fixed in fbb4dae by running bail_if_exited on the successful tick too, so it reports the exit status and log tail. The residual microsecond window is inherent.

src/daemon.rs
maxgio92: Fixed in fbb4dae. The timeout path now stops the daemon before it bails: TERM to the process group, KILL to the group after 500ms, then a wait to reap. The exit branches reuse the same signal_group helper. A final server_alive check runs before the kill, since the last poll can be 500ms old. The error message names the stop and suggests running llmman serve in the foreground when the first llama-server download is slow.

maxgio92:
Addressing also the Copilot's points

ericcurtin (changes requested):
Re-reviewed the diff directly (not the bot summaries) after re-reading my own
earlier comment on this thread, and I need to walk part of it back: my
"the still-running daemon is never terminated on timeout" comment and
Copilot's *suppressed* comment on the same commit are in direct conflict, and
Copilot is right.

**Blocking issue:** the timeout path at the bottom of `ensure_server`
(`signal_group(child.id(), false)` / 500ms / `signal_group(..., true)` /
`child.wait()`, right before the `"did not start within 60s and was
stopped"` bail) unconditionally TERM+KILLs the daemon's whole process group
whenever it hasn't opened its port within 60s — including while it's in the
middle of an `llama-server` auto-download.

`http_client()` in `src/llama_release.rs` sets a 30-minute HTTP timeout
specifically because these downloads (Windows CUDA packages especially) can
legitimately run that long, the download path has no resume support, and it
writes to PID-specific staging files. So killing here doesn't just fail the
current `pull`/`run`/`launch` — it discards whatever was downloaded and every
retry starts from zero.

That's a functional regression introduced by *this* PR, not a pre-existing
one: before this change, `ensure_server` returned the 60s timeout error and
left the daemon running, so a slow first-time download would finish in the
background and the next command would just connect to it. After this change
it never gets the chance to.

To make this mergeable, one of:
- Don't signal the group on the plain timeout path at all — only
  `bail_if_exited`'s exited-child branch should do it, since there the
  daemon is already dead and it's just reaping an orphaned `llama-server`.
  Keep the old "leave it running" behavior for a live-but-slow daemon, just
  with the improved `log_tail` message.
- Or, if killing a slow-but-alive daemon on timeout is genuinely wanted,
  make it conditional on there being no download in progress (e.g. a state
  marker the daemon writes once a download starts), and add a test that lets
  60s elapse mid-"download" and asserts it's allowed to finish rather than
  being killed.

**Non-blocking, worth doing while this file is already open:**
- `src/daemon.rs` ~443-444: after the post-loop `server_alive()` succeeds
  there's no second `bail_if_exited` call, so a daemon whose leader exited in
  that exact window but left a process-group child (e.g. `llama-server`)
  holding the port would report `Ok(())` without ever being checked.
  CodeRabbit has flagged this twice now.
- The exited-branch signal in `bail_if_exited` (immediate
  `signal_group(pid, false)`, no sleep/KILL) and the timeout-path signal
  (TERM, 500ms grace, KILL) use different escalation — worth unifying so a
  slow-to-die child is reaped consistently in both places.
- `log_tail_missing_file_points_at_path` derives its temp path from
  `process::id()` alone with no cleanup before the call; a stale file left
  by a crashed prior run would make the missing-file assertion pass for the
  wrong reason. Cheap fix: best-effort `remove_file` before calling
  `log_tail`, same as the sibling test already does after.

None of this touches the core fix — retaining `Child` and polling
`try_wait()` in the loop is correct and the new unit/e2e tests for it read
well. Once the timeout-kill question above is settled (and ideally with a
test covering the "still downloading past 60s" case), this is good to merge.

maxgio92:
Thank you @ericcurtin. I'm prone to implement a state marker so to prevent from mid-run killing the daemon.

I can address the remaining points as well within this PR, of course. Thanks!

maxgio92:
@ericcurtin all four addressed in e15dff2, taking your second option for the timeout kill.

1. The daemon keeps a .downloading marker fresh while downloading; on timeout a fresh marker leaves it running, otherwise it is stopped as before. Unit-tested at the predicate level; no mid-download e2e, since nothing keeps a daemon deterministically alive past 60s without binding.
2. Post-loop probe now matches the in-loop success path.
3. One stop_group helper (TERM, 500ms grace, KILL) serves both paths.
4. The log_tail test removes the temp path before asserting.

Also rebased onto current main; the branch predated #294's merge.

ericcurtin (approved):
Re-reviewed the current head (e15dff2) against my earlier `CHANGES_REQUESTED`.

**The blocking issue is resolved.** e15dff2 does exactly what I asked for: a `.downloading` marker (touched every `PROGRESS_LOG_INTERVAL` from the download/extract loop, `Drop`-cleaned) lets `ensure_server`'s timeout path tell a genuinely slow first-time `llama-server` download apart from a hung/dead daemon. It only stops the process group when no download is in progress, with a unit-tested staleness check (`marker_is_fresh`) covering the crashed-writer case. The old "leave it running" behavior is preserved for the case that matters (an in-progress, non-resumable download), and the process group is still cleaned up on genuine timeout/exit, closing the leak I originally flagged in fbb4dae.

Also confirmed while re-reviewing:
- The post-loop success path now re-runs `bail_if_exited` before returning `Ok(())`, closing the narrow bind-then-die-during-init window (my other inline comment) — verified with the new `download_marker_*` and `marker_is_fresh_*` unit tests, all green locally.
- `bail_if_exited` now rechecks `server_alive()` before bailing, so the concurrent-`serve`-launchers race (loser exits `EADDRINUSE`, winner's daemon is up) no longer surfaces as a false failure.
- Process-group signaling is unified into one `stop_group` helper (TERM, 500ms grace, KILL) shared by the exited and timeout branches, output nulled on both Unix `kill` and Windows `taskkill`.
- The `log_tail` test no longer false-fails/passes on a temp path that happens to contain the test's own log content.
- New `tests/launch_e2e.rs::ensure_server_fails_fast_when_daemon_dies_at_startup` exercises the original fail-fast regression end-to-end against the real binary (dead daemon → reports exit status + log tail in well under 60s), closing the "core regression path isn't covered" gap.
- Locally: `cargo build --all-targets`, `cargo test --lib daemon:: llama_release::`, `cargo clippy --all-targets` (no new warnings vs. main), and `cargo fmt --all -- --check` all clean on e15dff2. CI is green across all OS/arch matrix legs.

Resolved the review threads above that this addresses. Approving and merging — left a couple of non-blocking follow-ups as a comment.

ericcurtin:
Merged. A few non-blocking follow-ups worth opening issues for while this is fresh:

1. **No direct test for the new "leave daemon running while downloading" branch.** `download_in_progress()` and the marker lifecycle are unit-tested in isolation (`marker_is_fresh_*`, `download_marker_creates_touches_and_removes_the_marker`), but nothing exercises `ensure_server`'s timeout path actually taking that branch (e.g. a fake slow "download" that holds the marker fresh past 60s, asserting the daemon is *not* killed and the "still downloading" message is returned). Worth an e2e test alongside `ensure_server_fails_fast_when_daemon_dies_at_startup`.

2. **Marker isn't touched during extraction.** The doc comment on `try_ensure_from_network` calls this out and accepts it, but for a large archive, extraction time beyond `DOWNLOAD_MARKER_STALE_AFTER` (60s) past the last download touch would make `download_in_progress()` go false mid-extract, and a concurrently-timing-out client would kill the daemon during extraction instead of during download. Cheap fix: touch the marker once more right before `extract(...)` and/or periodically during extraction if that ever gets slow enough to matter in practice.

3. **Concurrent-download marker race** (two `llmman serve` processes downloading at once, first finisher's `Drop` deletes the shared marker before the survivor's next touch recreates it) is called out as an accepted narrow window in the code comment — fine to leave, but worth keeping in mind if flaky "daemon killed mid-download" reports show up in the wild with multiple daemons/hosts sharing an install root.

4. **Broader subprocess regression test for `ensure_server`** (CodeRabbit flagged this a couple of rounds back): a test that drives the full loop — spawn, poll, exit — via an injectable/fake child rather than the real binary, to cover timing edges the e2e test can't hit deterministically (e.g. exit landing exactly between the last in-loop poll and the post-loop check).

None of these block anything that shipped; flagging so they don't get lost.

PR #292, opened 2026-08-26
Add LLMMAN_HOST to override the serve daemon's bind address

ericcurtin:
Adds `LLMMAN_HOST` to control the address `llmman serve` binds to.

- `daemon::SERVER` is replaced by three functions, parsed from `LLMMAN_HOST` once per process:
  - `bind_addr()` — raw `host:port` the listener binds (wildcard hosts kept as-is).
  - `server()` — `http://host:port` every client talks to; wildcard hosts rewritten to loopback, scheme forced to `http`.
  - `connect_addr()` — bare `host:port` for the liveness check, same loopback rewrite.
- Host classification (`connectable_host`/`is_local_host`) parses the IP by value, so expanded IPv6 forms are caught too.
- `ensure_server`'s stale-daemon replacement is skipped for a remote host, so a remote PID is never used for local process control.
- `server_alive` uses a short `connect_timeout` instead of an unbounded connect, since a remote host may not respond.
- Defaults to `127.0.0.1:17434` when unset — no behavior change.
- README documents the variable; unit tests cover the parser.

Tested: `cargo test --lib` (264 passed), `cargo clippy`/`cargo fmt --check` clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ Start the inference server. Uses `llama-server` from [llama.cpp](https://github.
 llmman serve
 ```
 
-The server listens on `127.0.0.1:17434` and exposes:
+The server listens on `127.0.0.1:17434` by default, overridable via `LLMMAN_HOST`, and exposes:
 
 | API | Endpoints |
 |-----|-----------|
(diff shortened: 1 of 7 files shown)


ericcurtin:
@coderabbitai review

PR #294, opened 2026-08-26
fix(llama-release): follow llama.cpp's nightly-tag.txt pointer

### LINKED ISSUE #293: llama-server download broken by llama.cpp pointer releases
## Summary

The llama-server auto-download fails on every host without a `llama-server` on PATH, so `llmman serve` cannot start. llama.cpp [changed its release scheme](https://github.com/ggml-org/llama.cpp/releases) on 2026-08-25 and the resolver no longer finds any binary asset.

## Reproduction

With no `llama-server` on PATH:

    llmman serve

    Error: no llama-server on PATH and automatic download failed: no vulkan
    llama.cpp release asset found in v0.3.0 (looked for a name containing
    "-bin-ubuntu-vulkan-x64.tar.gz")

## Root cause

[`fetch_release`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/llama_release.rs#L67-L82) resolves [`releases/latest`](https://api.github.com/repos/ggml-org/llama.cpp/releases/latest) on `ggml-org/llama.cpp` when no version is pinned. llama.cpp now marks all `b<number>` binary releases as prereleases, which `releases/latest` skips, and publishes stable semver releases (currently [`v0.3.0`](https://github.com/ggml-org/llama.cpp/releases/tag/v0.3.0)) whose only asset is [`nightly-tag.txt`](https://github.com/ggml-org/llama.cpp/releases/download/v0.3.0/nightly-tag.txt): a one-line file naming the promoted binary build (currently `b10621`). Resolution therefore lands on a release with zero binaries and [`find_asset`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/llama_release.rs#L84-L89) matches nothing. The asset naming itself is unchanged: [`b10621`](https://github.com/ggml-org/llama.cpp/releases/tag/b10621) still carries `llama-b10621-bin-ubuntu-vulkan-x64.tar.gz`.

## Workarounds

Put a `llama-server` on PATH, run with `--ociman`, or pin a binary tag: `llmman serve --llama-cpp-version b10621 --pull-bin`.

maxgio92:
Closes #293.

llama.cpp marks its [`b<number>` binary releases](https://github.com/ggml-org/llama.cpp/releases) as prereleases and publishes stable semver releases ([`v0.3.0`](https://github.com/ggml-org/llama.cpp/releases/tag/v0.3.0)) whose only asset is [`nightly-tag.txt`](https://github.com/ggml-org/llama.cpp/releases/download/v0.3.0/nightly-tag.txt), a one-line file naming the promoted `b<number>` build. GitHub's `releases/latest` skips prereleases, so unpinned resolution in [`fetch_release`](https://github.com/llmmanorg/llmman/blob/721fc9eba5b6e785a4f9846431802bd3791d2b33/src/llama_release.rs#L67-L82) landed on `v0.3.0`, found no binary asset, and the llama-server auto-download failed on every host without one on PATH.

Changes, all in `src/llama_release.rs`:

- New `resolve_release` dereferences the pointer after any fetch (latest or pinned): download `nightly-tag.txt`, read the `b<number>` tag, fetch that release, and pick the host asset from there. This tracks llama.cpp's promoted stable build instead of the newest CI build.
- One hop only: a dereferenced release that is itself just a pointer errors out, naming both tags. Failed pointer downloads and failed tag fetches also name both tags.
- A release carrying binaries alongside a pointer file is used as-is; the pointer must be the sole asset to trigger dereferencing. Pinned `b<number>` tags behave unchanged.
- Doc note: pinning a pointer tag (`v0.3.0`) re-resolves over the network each run since the cache keys on the dereferenced `b<number>` tag; only a `b<number>` pin is reproducible.

## Testing

- 2 new unit tests (pointer detection including a mixed pointer-plus-binaries release, tag trimming); full lib suite passes.
- Verified end to end on a Linux vulkan host: `llmman serve --pull-bin` resolved `v0.3.0` to [`b10621`](https://github.com/ggml-org/llama.cpp/releases/tag/b10621), downloaded `llama-b10621-bin-ubuntu-vulkan-x64.tar.gz`, and the daemon started and served requests.

### DIFF
--- a/src/llama_release.rs
+++ b/src/llama_release.rs
@@ -81,6 +81,76 @@ fn fetch_release(client: &reqwest::blocking::Client, version: Option<&str>) -> R
         .with_context(|| format!("parse release metadata from {url}"))
 }
 
+/// Name of the pointer asset llama.cpp attaches to its stable semver
+/// releases (e.g. `v0.3.0`): a one-line text file containing the tag of
+/// the promoted `b<number>` build that carries the actual binaries.
+const NIGHTLY_POINTER: &str = "nightly-tag.txt";
+
+/// Returns the pointer asset if `release` is one of llama.cpp's semver
+/// pointer releases rather than a `b<number>` binary release. The pointer
+/// must be the release's sole asset: a release carrying binaries alongside
+/// a pointer file is a binary release and must be used as-is, not
+/// dereferenced away (or refused) just because the pointer exists.
+fn pointer_asset(release: &Release) -> Option<&Asset> {
+    if release.assets.len() != 1 {
+        return None;
+    }
+    release.assets.iter().find(|a| a.name == NIGHTLY_POINTER)
+}
+
+/// Parses the tag out of a downloaded `nightly-tag.txt`: trims
+/// surrounding whitespace (the file ends in a newline) and rejects an
+/// empty result.
+fn parse_pointer_tag(content: &str) -> Option<&str> {
+    let trimmed = content.trim();
+    (!trimmed.is_empty()).then_some(trimmed)
+}
+
+/// Fetches release metadata like [`fetch_release`], then dereferences
+/// llama.cpp's new pointer scheme: since the `b<number>` binary releases
+/// were all marked prerelease, `releases/latest` resolves to a stable
+/// semver release (e.g. `v0.3.0`) whose only asset is [`NIGHTLY_POINTER`],
+/// naming the `b<number>` release that actually carries the binaries.
+/// One hop only: a pointer chain longer than that is a publishing bug
+/// upstream and gets reported rather than followed.
+fn resolve_release(client: &reqwest::blocking::Client, version: Option<&str>) -> Result<Release> {
+    let release = fetch_release(client, version)?;
+    let Some(pointer) = pointer_asset(&release) else {
+        return Ok(release);
+    };
+    let pointer_tag = &release.tag_name;
+    let content = fetch_text(client, &pointer.browser_download_url).with_context(|| {
+        format!("download {NIGHTLY_POINTER} from pointer release {pointer_tag}")
+    })?;
+    let tag = parse_pointer_tag(&content).ok_or_else(|| {
+        anyhow!("pointer release {pointer_tag}'s {NIGHTLY_POINTER} is empty")
+    })?;
+    let dereferenced = fetch_release(client, Some(tag)).with_context(|| {
+        format!("pointer release {pointer_tag} names tag {tag}, but fetching that release failed")
+    })?;
+    if pointer_asset(&dereferenced).is_some() {
+        anyhow::bail!(
+            "pointer release {pointer_tag} names tag {tag}, which is itself \
+             only a {NIGHTLY_POINTER} pointer release (refusing to follow a \
+             pointer chain)"
+        );
+    }
+    Ok(dereferenced)
+}
+
+/// Downloads `url` as plain text, for the small [`NIGHTLY_POINTER`] file.
+fn fetch_text(client: &reqwest::blocking::Client, url: &str) -> Result<String> {
+    let resp = client
+        .get(url)
+        .header("user-agent", "llmman")
+        .send()
+        .with_context(|| format!("query {url}"))?;
+    if !resp.status().is_success() {
+        anyhow::bail!("download {url} returned {}", resp.status());
+    }
+    resp.text().with_context(|| format!("read body of {url}"))
+}
+
 fn find_asset<'a>(release: &'a Release, must_contain: &str) -> Option<&'a Asset> {
     release
         .assets
@@ -428,6 +498,10 @@ pub struct Resolved {
 /// when given, fetches that exact release tag (e.g. "b10360") instead of
 /// whatever is currently latest — once cached under that tag it is never
 /// re-fetched, so pinning also makes this fully reproducible across runs.
+/// Only a `b<number>` pin gets that guarantee: pinning a semver pointer
+/// tag (e.g. "v0.3.0") re-resolves through the network each run, since the
+/// cache keys on the `b<number>` tag it dereferences to (see
+/// `resolve_release`) and the pointer file is mutable upstream.
 ///
 /// Blocking (network + disk I/O) — callers on an async runtime must run
 /// this via `tokio::task::spawn_blocking` (see `cmd::serve`'s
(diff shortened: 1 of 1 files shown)


ericcurtin:
We failed the build, we should fix that and any review comments worth fixing from coderabbit and copilot.

As a follow-up, we should remove "automatic download" feature. That wasn't intentionally added. Although manual downloads "--pull-bin", "--ociman" and "--pull-oci" were. Going to review the rest soon, because even still "--pull-bin" shouldn't fail.

maxgio92:
Thank you @ericcurtin, I'm going to take a look at them

ericcurtin:
@maxgio92 thank you! I think the rest of the PR is relatively ok. Although I'm getting Claude Sonnet to dump some comments now, again address any worth addressing.

ericcurtin (commented):
Reviewed the diff, checked the live GitHub API to confirm the premise, and ran the test/build/clippy suite against this branch.

**Confirmed the bug and the fix against the real API** (as of today):
- `GET /repos/ggml-org/llama.cpp/releases/latest` -> `v0.3.0`, `assets: ["nightly-tag.txt"]` — exactly the failure mode described.
- With this branch's `resolve_release`, that correctly dereferences to `b10621`'s real asset list.

**Verification performed:**
- `cargo build --lib` — clean.
- `cargo test --lib llama_release` — all 8 non-ignored tests pass (`pointer_asset_detects_a_semver_pointer_release`, `parse_pointer_tag_trims_and_rejects_blank_content`, etc.).
- `cargo clippy --lib --all-targets` — no new warnings from this diff.

Code-reading notes: the "one hop only" guard, the sole-asset check in `pointer_asset` (so a real binary release that happens to also carry a `nightly-tag.txt` isn't mistaken for a pointer), and the pinned-tag cache-key docs all check out against `try_ensure_from_network`/`ensure_llama_server`'s call sites. Left a couple of small non-blocking notes inline.

Overall this looks correct and well-tested; nice root-causing in the PR description.

src/llama_release.rs
```diff
@@ -81,6 +81,76 @@ fn fetch_release(client: &reqwest::blocking::Client, version: Option<&str>) -> R
...
+/// Returns the pointer asset if `release` is one of llama.cpp's semver
+/// pointer releases rather than a `b<number>` binary release. The pointer
+/// must be the release's sole asset: a release carrying binaries alongside
+/// a pointer file is a binary release and must be used as-is, not
+/// dereferenced away (or refused) just because the pointer exists.
+fn pointer_asset(release: &Release) -> Option<&Asset> {
+    if release.assets.len() != 1 {
+        return None;
+    }
+    release.assets.iter().find(|a| a.name == NIGHTLY_POINTER)
```
ericcurtin, line 98: Heuristic note (non-blocking): this keys purely off "exactly one asset, named `nightly-tag.txt`". I checked the actual release history and llama.cpp hasn't always been this consistent — `v0.1.2` (an older prerelease) shipped a pointer-like file named `nightly.tag` (no dash, different extension) *alongside* a UI tarball, i.e. 3 assets, none matching this exact name. That tag won't hit this code path today (multiple assets -> `pointer_asset` returns `None` -> falls through to the normal `find_asset` flow, which will just fail to find an OS asset with a slightly worse error message), so it's not a regression, just a reminder that this is coupled to today's exact publishing convention for `v0.3.0`/`v0.2.0` and may need a follow-up if upstream tweaks the pointer filename or bundles a checksum file alongside it again.

src/llama_release.rs
```diff
@@ -81,6 +81,76 @@ fn fetch_release(client: &reqwest::blocking::Client, version: Option<&str>) -> R
...
+            "pointer release {pointer_tag} names tag {tag}, which is itself \
+             only a {NIGHTLY_POINTER} pointer release (refusing to follow a \
+             pointer chain)"
+        );
+    }
+    Ok(dereferenced)
+}
+
+/// Downloads `url` as plain text, for the small [`NIGHTLY_POINTER`] file.
+fn fetch_text(client: &reqwest::blocking::Client, url: &str) -> Result<String> {
```
ericcurtin, line 142: Minor nit: `fetch_text` duplicates the request-building/status-check boilerplate from `fetch_release` (user-agent header, `send()`, `is_success()` check). Could factor the common "GET url, check status" part into a small shared helper that both `fetch_release` (before `.json()`) and this (before `.text()`) call — not essential, just a small dedup opportunity now that there are two near-identical GET helpers in the file.

src/llama_release.rs
```diff
@@ -428,6 +498,10 @@ pub struct Resolved {
 /// when given, fetches that exact release tag (e.g. "b10360") instead of
 /// whatever is currently latest — once cached under that tag it is never
 /// re-fetched, so pinning also makes this fully reproducible across runs.
+/// Only a `b<number>` pin gets that guarantee: pinning a semver pointer
+/// tag (e.g. "v0.3.0") re-resolves through the network each run, since the
```
ericcurtin, line 502: Just confirming my understanding (matches the doc comment above at L501-504, no change requested): when `pinned_version` is a semver pointer tag like `v0.3.0`, this fast-path lookup uses the *raw* pinned tag for `install_dir`, which never matches the cache directory (keyed on the dereferenced `b<number>` tag), so a `v0.3.0`-pinned run always falls through to `try_ensure_from_network` and re-hits the network for the two metadata calls (pointer release + dereferenced release) even when the binary is already on disk from a prior run. Functionally fine (no re-download, and it's documented), just flagging in case a fully-offline pinned-pointer path is ever wanted later — would need to persist the tag->dereferenced-tag mapping locally.

src/llama_release.rs
maxgio92: @ericcurtin Thank you. Yeah, that was my concern indeed as we couple with their release strategy and conventions which may mutate too. My istinct is that long-term the download is a user responsibility so that a coupling can be removed. WDYT?

src/llama_release.rs
maxgio92: @ericcurtin That's correct. I think it's something I can improve to make the local cache paths consistent.
I am thinking about either solving it on the write- or on the read-path. So on the read-path we could rely on the immutable digest, while on the write-path we could just be consistent with what's expected by the read path, so relying on the release tag, ignoring its mutability (so either we go consistent relying on digets or on release tags like now). WDYT?

PR #303, opened 2026-08-26
fix: bound metadata fetch and validate refs

### LINKED ISSUE #302: Malformed model references retry for a minute
`pull`/`run`/`launch` accept any string as a model reference and forward it to the network layer without validation, so a structurally broken reference is not rejected but retried 8 times with exponential backoff, hanging the command for over a minute.

## Reproduction (all hang ~60s+, exit 124)

    llmman pull ../../../etc/passwd
    llmman pull oci://
    llmman pull 'http://evil.test/x'

serve.log shows the reference spliced straight into a URL, then the retry loop:

    GET https://../api/models/../../etc/passwd  ...  invalid dns name
    GET https://oci:/api/models//  ...  lookup oci: no such host
    retrying ... (attempt 6/8, wait 18.6s)

Related, same cause (a reference that survives resolution but can never resolve): a whitespace-only reference becomes `docker.io/ai/ ` and hits the registry, and a reference with a control character reaches URL parsing before failing (`invalid control character in URL`). A genuinely empty `pull ""` already fails fast with `model is required`; blank-after-trim and malformed references should too.

## Cause

[`resolve`](https://github.com/llmmanorg/llmman/blob/8de4ea39afe1e1f90fb5d07e8c5f8f527c17efe3/src/shortnames.rs#L120-L160) and [`resolve_ollama_api`](https://github.com/llmmanorg/llmman/blob/8de4ea39afe1e1f90fb5d07e8c5f8f527c17efe3/src/shortnames.rs#L194) pass any input through to a source handler with no shape check: `../../../etc/passwd` becomes `hf.co/../../../etc/passwd`, `oci://` and `http://...` reach the resolver as-is. The 8-attempt retry then treats an unresolvable host as a transient network error.

maxgio92:
Closes #302. Reshaped per review: instead of classifying transport errors (fragile, and asymmetric between the Rust HF client and the Go path, as @ericcurtin caught), this bounds the cheap request and validates refs up front.

A malformed or unresolvable reference used to run the 8-attempt (~127s) blob-download retry budget on a cheap metadata request before failing. Two changes:

- Bound the metadata GET/HEAD to a single attempt on both the Rust HF client and the Go HF path, so any bad host fails at once regardless of error type. Blob downloads and pushes keep the full budget. The OCI image-copy path wraps manifest and blobs in one retry, so its manifest phase is a documented residual (a typo'd but syntactically valid OCI registry host); malformed refs are rejected before reaching it.
- Add `validate_reference`, an up-front gate at the `pull`/`run`/`launch` entry points and the daemon pull handler, rejecting blank, control-char, unknown-scheme, `..`, and percent-encoded refs before any network I/O.

Both retry classifiers revert to 4xx-only; the dead `retry` helper and the direct `http` dependency are removed.

## Testing

Unit tests for `validate_reference` (accepts bare/owner-repo/host-owner-repo with port/tag/@digest, all known schemes, local paths; rejects the garbage classes) and a single-attempt metadata test. `cargo fmt`/`clippy`/`test` and `go test ./...` pass. Verified end to end: the three #302 repros and `hf.coo/microsoft/phi-2` (typo host, the NXDOMAIN case) now fail in under 2s instead of ~127s.

Follow-ups from the review: a fuzz target for `validate_reference` (no valid parse yields a `..` or over-length part), and bounding the OCI copy path's manifest phase.

### DIFF
--- a/src/hf/client.rs
+++ b/src/hf/client.rs
@@ -20,7 +20,7 @@ const RETRY_BASE: Duration = Duration::from_secs(1);
 /// No bytes at all for this long aborts the current attempt.
 const STALL_TIMEOUT: Duration = Duration::from_secs(60);
 
-/// A non-2xx HTTP response, carrying enough structure for [`retry`] to
+/// A non-2xx HTTP response, carrying enough structure for a retry loop to
 /// react to more than just "was this permanent": a server-supplied
 /// `Retry-After` (RFC 9110 §10.2.3), mirroring `huggingface_hub`'s own
 /// handling of it on a 429.
@@ -49,7 +49,7 @@ impl HttpStatusError {
     }
 }
 
-/// How long [`retry`] will ever honor a server-supplied `Retry-After`
+/// How long a retry loop will ever honor a server-supplied `Retry-After`
 /// for. `huggingface_hub` trusts it completely; an unbounded wait here
 /// could eat a whole CI job's time budget on one file, so this caps it
 /// well short of that instead.
@@ -116,46 +116,21 @@ fn rand_unit() -> f64 {
     (nanos % 1_000_000) as f64 / 1_000_000.0
 }
 
-/// Calls `attempt` up to [`MAX_ATTEMPTS`] times with exponential backoff
-/// — or the previous attempt's `Retry-After`, if it had one — stopping
-/// immediately once the most recent error is [`is_permanent`]. Every
-/// attempt restarts its work entirely from scratch (mirrors
-/// `retryStream`'s own doc comment: there's no partial state to resume
-/// into on this side of a registry push either way).
-pub async fn retry<T, F, Fut>(label: &str, mut attempt: F) -> Result<T>
+/// Runs `attempt` exactly once, with no retry loop or backoff. The cheap
+/// metadata requests (the model-info GET and the HEAD size/etag probes)
+/// use this so a bad or nonexistent host fails immediately rather than
+/// running a full backoff budget. The HF client attaches its bearer token
+/// up front (no 401 auth handshake to retry), and the per-request timeout
+/// reqwest already applies still bounds a slow-but-resolvable host. A 4xx
+/// surfaces as-is.
+pub async fn once<T, F, Fut>(label: &str, attempt: F) -> Result<T>
 where
-    F: FnMut() -> Fut,
+    F: FnOnce() -> Fut,
     Fut: std::future::Future<Output = Result<T>>,
 {
-    let mut last_err: Option<anyhow::Error> = None;
-    for i in 0..MAX_ATTEMPTS {
-        if i > 0 {
-            let delay = last_err
-                .as_ref()
-                .and_then(retry_after_of)
-                .unwrap_or_else(|| retry_delay(i));
-            eprintln!(
-                "\n[llmman] retrying {label} (attempt {}/{MAX_ATTEMPTS}, wait {delay:?})",
-                i + 1
-            );
-            tokio::time::sleep(delay).await;
-        }
-        match attempt().await {
-            Ok(v) => return Ok(v),
-            Err(e) => {
-                let permanent = is_permanent(&e);
-                eprintln!("[llmman] {label} error: {e:#}");
-                last_err = Some(e);
-                if permanent {
-                    break;
-                }
-            }
-        }
-    }
-    Err(anyhow!(
-        "{label} failed after {MAX_ATTEMPTS} attempts: {:#}",
-        last_err.expect("loop always runs at least once")
-    ))
+    attempt().await.inspect_err(|e| {
+        eprintln!("[llmman] {label} error: {e:#}");
+    })
 }
 
 // ---------------------------------------------------------------------------
@@ -310,70 +285,15 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn retry_stops_immediately_on_permanent_error() {
+    async fn once_runs_the_attempt_a_single_time_on_failure() {
         let attempts = AtomicU32::new(0);
-        let result: Result<()> = retry("test", || {
+        let result: Result<()> = once("test", || {
             attempts.fetch_add(1, Ordering::SeqCst);
-            async {
-                Err(anyhow::Error::new(HttpStatusError::new(
-                    "GET x",
-                    404,
-                    &HeaderMap::new(),
-                )))
-            }
+            async { Err(anyhow!("boom")) }
         })
         .await;
         assert!(result.is_err());
         assert_eq!(attempts.load(Ordering::SeqCst), 1);
-    }
-
-    #[tokio::test]
-    async fn retry_succeeds_after_transient_failures() {
-        let attempts = AtomicU32::new(0);
-        let result = retry("test", || {
-            let n = attempts.fetch_add(1, Ordering::SeqCst);
-            async move {
-                if n < 2 {
-                    Err(anyhow!("transient"))
-                } else {
-                    Ok(42)
-                }
-            }
-        })
-        .await
-        .unwrap();
-        assert_eq!(result, 42);
-        assert_eq!(attempts.load(Ordering::SeqCst), 3);
-    }
-
-    #[tokio::test]
-    async fn retry_honors_retry_after_over_exponential_backoff() {
-        let attempts = AtomicU32::new(0);
-        let start = Instant::now();
-        retry::<(), _, _>("test", || {
-            let n = attempts.fetch_add(1, Ordering::SeqCst);
-            async move {
-                if n == 0 {
-                    Err(anyhow::Error::new(HttpStatusError {
-                        prefix: "GET x".to_string(),
-                        status: 429,
-                        retry_after: Some(Duration::from_millis(100)),
-                    }))
-                } else {
-                    Ok(())
-                }
-            }
-        })
-        .await
-        .unwrap();
-        // Default exponential backoff for the first retry is ~0.75-1.25s;
-        // a Retry-After of 100ms is a clearly distinguishable, much
-        // shorter wait if actually honored.
-        assert!(
-            start.elapsed() < Duration::from_millis(500),
-            "elapsed={:?}",
-            start.elapsed()
-        );
     }
 
     #[tokio::test]
--- a/go-shim/backend_podman.go
+++ b/go-shim/backend_podman.go
@@ -370,6 +370,13 @@ func copyImageWithProgress(ctx context.Context, pctx *signature.PolicyContext, d
 	copyImageMu.Lock()
 	defer copyImageMu.Unlock()
 
+	// copy.Image fetches the manifest and every blob inside this one
+	// retryStream, so the manifest phase is not separately bounded to a
+	// single attempt the way the Rust HF metadata GET is (hf::client::once). A
+	// typo'd but syntactically valid OCI registry host can still run the
+	// full retry budget on its manifest lookup here: a known residual, left
+	// as a follow-up. Malformed refs never reach this path: validate_reference
+	// (Rust, Layer B) rejects them up front.
 	var manifestData []byte
 	err := retryStream(ctx, progressKey, isHTTP4xx, func() error {
 		data, err := copyImageAttempt(ctx, pctx, dst, src, present, pastTense, opts, changed, progressKey)
--- a/go-shim/is_http4xx_test.go
+++ b/go-shim/is_http4xx_test.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestIsHTTP4xx(t *testing.T) {
+	cases := []struct {
+		name string
+		err  error
+		want bool
+	}{
+		{"nil is not permanent", nil, false},
+		{"HTTP 400 substring is permanent", errors.New("GET https://example.com/f: HTTP 400"), true},
+		{"HTTP 401 substring is permanent", errors.New("GET https://example.com/f: HTTP 401"), true},
+		{"HTTP 403 substring is permanent", errors.New("GET https://example.com/f: HTTP 403"), true},
+		{"HTTP 404 substring is permanent", errors.New("GET https://example.com/f: HTTP 404"), true},
+		{"HTTP 500 stays retryable", errors.New("GET https://example.com/f: HTTP 500"), false},
+		{"plain error stays retryable", errors.New("connection refused"), false},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got := isHTTP4xx(c.err); got != c.want {
+				t.Errorf("isHTTP4xx(%v) = %v, want %v", c.err, got, c.want)
+			}
+		})
+	}
+}
--- a/src/bin/llmman-bench.rs
+++ b/src/bin/llmman-bench.rs
@@ -170,7 +170,7 @@ fn run(args: &Args) -> anyhow::Result<()> {
 
     let mut results = Vec::with_capacity(args.model.len());
     for raw_model in &args.model {
-        let model = shortnames::resolve_ollama_api(raw_model);
+        let model = shortnames::resolve_ollama_api(raw_model)?;
         // Fail fast on a bad/unresolvable reference rather than only
         // discovering it partway through warmup below.
         daemon::ensure_model_pulled(&model)?;
--- a/src/cmd/build.rs
+++ b/src/cmd/build.rs
@@ -22,6 +22,11 @@ pub struct BuildArgs {
 }
 
 pub fn run(args: &BuildArgs) -> anyhow::Result<()> {
+    // The build tag is a user-chosen label stored verbatim, not resolved, so
+    // this is its only validation gate. Check it before touching the store so
+    // a bad tag never creates the store tree.
+    crate::shortnames::validate_reference(&args.tag)?;
+
     let store_root = crate::default_store()?;
     let store = OciStore::open(&store_root)?;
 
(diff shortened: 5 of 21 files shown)


ericcurtin:
If we fix the conflicts we can trigger the builds, haven't reviewed yet, will take a look tomorrow.

src/hf/client.rs
```diff
@@ -80,12 +80,65 @@ fn is_permanent_status(status: u16) -> bool {
...
+
+/// True for the reqwest errors that a retry can never fix: a builder error
+/// (the URL/name could not be turned into a request at all) or a connect
+/// error rooted in an invalid URI or invalid DNS name. A generic connect
+/// failure (refused, unreachable, timed out) stays retryable.
+fn is_permanent_reqwest(err: &reqwest::Error) -> bool {
+    if err.is_builder() {
+        return true;
+    }
+    err.is_connect() && has_invalid_target(err)
```
ericcurtin: Go treats NXDOMAIN as permanent; here it's a plain `is_connect()` with a `dns error` source, so it falls through to retryable:

```
PROBE https://evil.test/x
  source[1] = dns error
  => is_permanent = false
```

Because `hf::classify` only routes `://` refs to Go, a typo'd host lands here and still takes ~127s.

Revising my earlier suggestion: I'd close this by not retrying the metadata GET at all, rather than by mirroring the DNS classification into Rust. Fixing the budget fixes every unclassified error at once. See the review body.

src/hf/client.rs
```diff
@@ -80,12 +80,65 @@ fn is_permanent_status(status: u16) -> bool {
...
+}
+
+/// Walks a reqwest error's source chain looking for a host that can never
+/// resolve because it is syntactically invalid.
+fn has_invalid_target(err: &reqwest::Error) -> bool {
+    use std::error::Error;
+    let mut src = err.source();
+    while let Some(e) = src {
+        if e.downcast_ref::<http::uri::InvalidUri>().is_some() {
```
ericcurtin: Is this reachable? `has_invalid_target` only runs when `is_connect()`, but a URI that fails to parse surfaces as a *builder* error, which the line above already short-circuits. In my probe of the #302 repro the chain was `client error (Connect)` -> `invalid dns name`; the `InvalidUri` downcast never fired.

If it's dead, dropping it also removes the need for the new direct `http` dependency.

src/hf/client.rs
```diff
@@ -80,12 +80,65 @@ fn is_permanent_status(status: u16) -> bool {
...
+            return true;
+        }
+        // rustls's invalid-DNS-name error type is not a direct dependency,
+        // so match its Display form as a last resort. This is narrower than
+        // `is_connect()` alone, which also covers transient failures we
+        // must keep retrying.
+        if e.to_string()
```
ericcurtin: This is the branch that actually fixes `pull ../../../etc/passwd`, and it's the only new branch with no test, and it rests on rustls's `InvalidDnsNameError` Display staying exactly `"invalid dns name"`.

Rather than pin it with a test, I'd delete it. An up-front character allowlist on the reference rejects `..` before any DNS lookup happens, which is where a path-traversal shape belongs. Catching it by inspecting a TLS error string is compensating at the wrong layer.

go-shim/shared_oci.go
```diff
@@ -460,11 +462,33 @@ func retryAfter(err error) (time.Duration, bool) {
...
+// retried.
+func isPermanent(err error) bool {
 	if err == nil {
 		return false
 	}
+
+	// NXDOMAIN: the host name does not exist, so no retry will resolve it.
+	// A timed-out or temporary DNS failure stays retryable.
+	var dnsErr *net.DNSError
+	if errors.As(err, &dnsErr) && dnsErr.IsNotFound {
```
ericcurtin: Classification is correct: `errors.As` does reach through the production `*url.Error{Op:"Get"}` wrapper, verified against a live resolver.

But NXDOMAIN isn't reliably permanent: resolvers return it transiently while a VPN comes up, across a captive portal, or while cluster DNS restarts, which is exactly when a long pull tends to be started. Bounding the metadata fetch's retry budget gets the same latency win without giving up that recovery, and without needing this branch at all.

ericcurtin (commented):
Checked out the branch; both suites pass (`cargo test --lib hf::client` 15/15, `go test -run TestIsPermanent`). Verified the three repros in #302 against real error shapes. It fixes all three, but two are fixed by the Go classifier and one by the Rust one, and that asymmetry leaves a gap.

## Gap: Rust doesn't treat NXDOMAIN as permanent, Go does

`hf::classify` (src/hf/mod.rs:298) only routes a ref to Go when it contains `://` or starts with `/`; everything else fails closed to the Rust path. So the scheme-less neighbour of repro 3 is still unfixed:

```
ROUTE "evil.test/x" -> RUST(hf)
PROBE https://evil.test/x  => is_permanent = false   # 8 attempts, ~127s
```

A typo'd host (`llmman pull hf.coo/microsoft/phi-2`) is the likeliest real-world trigger, and it still burns the full backoff.

## Suggested direction (better, I think, than closing that gap as written)

**1. Don't retry the metadata GET.** The ~127s exists only because `/api/models/...` shares the 8-attempt budget sized for blob downloads. Give it a single attempt (plus an auth-challenge retry) or a short wall-clock deadline, and keep the generous budget for blobs. A bad host then fails in milliseconds regardless of how the error is classified, and most of the enumeration in this PR stops being necessary.

**2. Validate the reference before any network I/O.** This is your Layer B, and I'd land it first. Split into host/namespace/model/tag and enforce a strict byte allowlist per part: first byte `[A-Za-z0-9_]`, remainder `[A-Za-z0-9_.-]`, `:` only in host; length caps (host <= 350, others <= 80); substitute a sentinel for empty halves so a dangling separator can never validate. That rejects `../../../etc/passwd`, `oci://`, whitespace, control characters and percent-encoding with one instant, readable error, and makes path traversal structurally impossible instead of an accident of DNS resolution. Worth a fuzz target asserting no valid parse ever yields a `..` part or an over-long one.

**3. Prefer bounded retries over classified errors.** Keep 4xx -> permanent (structural and cheap) and keep the stop list tiny: cancellation, out of disk. Growing a transport-error denylist means every miss is another two-minute hang and another PR, while every false positive kills a legitimate transfer. Concretely, I'd drop the `"invalid dns name"` string match and the NXDOMAIN branch rather than mirror them across both languages.

## Smaller

- Dual `reqwest` in the tree (0.12 direct, 0.13 via `hf-xet`) means `downcast_ref::<reqwest::Error>()` silently misses xet-path errors, which do reach `is_permanent` via pull.rs:277.
- New direct `http = "1"` unifies with reqwest's today, but a future `cargo update` can split them, at which point the downcast stops matching with no compile or test failure.
- Nit: `retryStream`'s parameter is now named `isPermanent`, shadowing the package-level function (shared_oci.go:512).

go-shim/shared_oci.go
maxgio92: Right, a tinier budget dedicated to the metadata GET would achieve it without compromising the expected transient NXDOMAIN scenarios, thanks.

src/hf/client.rs
maxgio92: Thanks, I agree, I'm going to implement an upper layer with allowlists like planned for the layer B in #302. Also, I think it especially makes sense considering the asymmetry that I didn't catch between the Rust hugging face client and the Go OCI/registries ones

src/hf/client.rs
maxgio92: My bad, you're right, it is

src/hf/client.rs
maxgio92: > Revising my earlier suggestion: I'd close this by not retrying the metadata GET at all

and

> Give it a single attempt (plus an auth-challenge retry)

(from the first point of the review)

Just to be on the same page with the single attempt for metadata GETs and the authn challenge: the HF client attaches the token up front and has no 401 auth-challenge handshake on this path (that's the OCI bearer flow on the Go side), and a 401 is already permanent. So here a single attempt means literally one request if I'm not mistaken, no auth retry to add. Did you mean the auth-challenge retry for the OCI path, or is there an HF re-auth case I'm missing?

maxgio92:
Thank you @ericcurtin for the thorough review. My bad I didn't catch the asymmetry and that changes the plan, and your direction makes sense to me.

I'm going to reshape this PR to implement the layer B directly as well as addressing the metadata GET tiny budget (single attempt for now, happy to add the auth-challenge and/or a wall-clock deadline in case, see the [comment](https://github.com/llmmanorg/llmman/pull/303#discussion_r3869054100)), and dropping the transport-error classification in favor of the simpler approach above.

On Layer B: the refs we accept are all over the place (bare names, owner/repo, host/owner/repo with port/tag/`@digest`, the hf/ms/ngc/s3/gs schemes, local paths). A strict per-part allowlist has to model every one of those, and the day it gets one wrong it rejects a valid ref, so I'd rather reject the obvious garbage (blank, control chars, unknown scheme, `..`, percent-encoding) and leave the rest alone. WDYT?

ericcurtin (changes requested):
Two gaps flagged are still present in `ff0151b`:
- `go-shim/shared_oci.go`'s `isPermanent` only catches `*url.Error` when `Op == "parse"`. Hostless/bad-scheme URLs surface via `Client.Do` with `Op == "Get"` and still hit the full 8-attempt backoff (CodeRabbit's first actionable comment, unresolved).
- `src/hf/client.rs`'s `has_invalid_target` only matches rustls's `"invalid dns name"` string. A real OS NXDOMAIN (the scheme-less repro through the Rust path) isn't caught and stays retryable (CodeRabbit's second actionable comment, unresolved).

A least one of the three repros in #302 is still not fixed.

We need to pick one path:
1. Reshape per the thread consensus, likely making most of this classifier work unnecessary.
2. If landing this fix first: close the two gaps with regression tests, fix the dual `reqwest` 0.12/0.13 downcast miss, and rename `retryStream`'s `isPermanent` param so it stops shadowing the package function.

Follow-ups either way:
- [ ] Fuzz test for Layer B: no valid parse should yield a `..` part or an over-length one.
- [ ] Pin `http`/`reqwest` versions so a future `cargo update` can't silently split them.

maxgio92:
Thanks Eric. I'd go with the first option, reshaping this PR as for the consensus

ericcurtin (changes requested):
Pulled `3ea9964` and ran the full suite: `cargo test --lib` (53 hf:: tests + 10 shortnames:: tests pass), `cargo clippy --all-targets` (no new warnings vs main), `cargo fmt --check` clean, `go build`/`go vet`/`go test ./...` clean. This is the reshape we discussed and it's a much better shape: single-attempt metadata fetch instead of a growing transport-error denylist, `validate_reference` as an up-front reject-list, both classifiers back to plain 4xx, the `retryStream` param-shadowing nit fixed, and the direct `http`/`url` deps dropped. Good direction, not quite there yet.

Two things I want fixed before merge:

1. `validate_reference` is only wired into `cmd::pull`/`run`/`launch` and the `/api/pull` handler. `ensure_model` (`serve.rs:2423`) is the shared boundary that `/api/generate`, `/api/chat`, `/api/chat/completions`, and the Anthropic-compat endpoint all call with a client-supplied `model` string (6 call sites, `serve.rs:3958,4051,4170,4386,4583,4951`), plus `ServeArgs::model`'s startup preload, and none of them validate first. A garbage ref sent to `/api/chat` skips the fast reject and goes straight to `resolve_ollama_api` + pull. Worse, if it happens to classify as an OCI host, it hits the manifest-phase residual you already flagged in `backend_podman.go` (one `retryStream` around manifest+blobs, not bounded to a single attempt), which is close to the original ~127s hang, just reachable from a different endpoint than the one this PR set out to fix. Move the `validate_reference` call into `ensure_model` itself (or wherever the pull is triggered from inside it) so every entry point gets it once, and drop the now-redundant check in the `/api/pull` handler.

2. CodeRabbit's catch on `src/shortnames.rs` is real: `hf:///foo` has `rest = "/foo"` after the `hf://` split, which is non-empty so `validate_reference` accepts it, but it's an empty-authority form. `resolve("hf:///foo")` produces `"hf.co//foo"` (verified locally). Reject known-scheme refs whose `rest` starts with `/`, with a regression test.

Minor, can be same commit or a fast follow-up: the doc comment on `once` (`client.rs`) still says "up to MAX_ATTEMPTS" and links to the now-deleted `retry` function, Copilot flagged this too.

Once (1) and (2) are in, this looks mergeable to me.

maxgio92:
@ericcurtin reshaped and pushed (8378e68). Your three blockers are addressed:

1. Validation moved to the shared boundary: `resolve`/`resolve_ollama_api` now validate up front, so `ensure_model` covers `/api/chat`, `/api/generate`, the OpenAI and Anthropic endpoints, and the `--model` preload. The redundant `/api/pull` handler check is gone; `InvalidReference` maps to 400 centrally.
2. `hf:///foo` and the resolved `hf.co//foo` are both rejected via an empty-path-component rule, with a resolver-level test.
3. `once` doc comment fixed (plus two other stale `retry` doc links).

On Layer B's shape: I went with a kind-dispatched reject-list rather than the per-part allowlist, so object-store keys (`s3`/`gs`/`ngc`) stay opaque and only registry/HF refs get the strict path rules, no per-scheme special-casing. Happy to switch to the allowlist if you'd rather.

Also reverted both classifiers to 4xx-only and dropped the direct `http` dep (which moots the version-split note). Two follow-ups from your list are still open, both deferrable: a fuzz test for `validate_reference` (with an over-length component cap to anchor it), and pinning `reqwest`; I can file those as separate issues. Ready for another look.

ericcurtin (changes requested):
Direction is right, both earlier blockers are fixed, and `cargo test`/`clippy` plus `go vet`/`test` are clean locally.

Blocking:

- Conflicts with `main`, needs a rebase.
- The scheme gate only matches `://`. On this branch `validate_reference("http:/evil.test/x")` and `("oci:/registry/ns/model")` both return `Ok`, then `has_host` reads `http:` as a host:port and probes it. Reject a `:/` prefix that isn't a known `scheme://`.

Follow-ups: parse into host/namespace/model/tag/digest with a per-part byte allowlist instead of a reject-list; retry metadata once on a 401 challenge; bound the OCI manifest phase.

maxgio92:
Thanks @ericcurtin! Fixed and rebased (cf16de5).

- Scheme gate: anything whose first `:` is followed by `/` is now rejected unless it's a known `scheme://`, so `http:/evil.test/x`, `oci:/...`, and junk like `1a://x` are all out. No real reference has `/` after its first colon, so nothing legitimate is caught. Tests pin the forms.
- Rebased onto main. Note #325 removed the Go HF fetch path, so the Go side of this PR shrank to just the classifier test; the single-attempt metadata bound now lives only in the Rust client.

The follow-ups (allowlist parse, 401 retry, OCI manifest bound) I'll file as separate issues.

ericcurtin (changes requested):
Build is red on every Linux/macOS target:

```
./backend_podman.go:387:39: undefined: isPermanent
```

`isPermanent` is only `retryStream`'s parameter name, never a package-level function. That call site is `isHTTP4xx` on `main`. Revert the one-line change (keep the new comment) and rename `is_permanent_test.go` to match the function it actually tests.

Conflicts with `main` again, `src/shortnames.rs` only: comment wording ("Go shim" -> `crate::sources`) plus `PASSTHROUGH_SCHEMES` vs the inline literal. Take your side of both.

Still the follow-up from last round, and I'd rather have it here than after merge: `validate_reference` is a reject-list over the raw string, so each new bad shape needs a new rule (`%`, `?`, `#`, `:/`, `..`, empty component, and the next one). Split the reference into host / namespace / model / tag / digest and validate each part against a byte allowlist with length bounds: first byte `[A-Za-z0-9_]`, rest `[A-Za-z0-9_.-]`, `:` only in host and digest, `.` not in namespace, host <= 350 bytes and the rest <= 80. Malformed input then fails by construction instead of by enumeration, and `has_host` gets a parsed host instead of guessing from a `:`.

The single-attempt metadata bound and the full budget on blobs are the right split, keep that.

maxgio92:
My bad @ericcurtin, I've missed the podman build flag. Gotcha, I've shaped it initially over less impactful deny-based list, but I agree that the allow list will be more correct and maintainable. I'm going to reshape it, thank you

maxgio92:
One check on the digest bound: the 80-byte cap rejects sha512 digests (135 bytes; OCI-legal but rare), sha256 fits. Should we keep it, or give the digest its own ~160-byte bound?

ericcurtin:
`MAX_PART_LEN` of 80 on the model part rejects real repos. 4 of the top 500 GGUF repos by downloads exceed it:

```
hf.co/DavidAU/Qwen3.6-40B-Claude-4.6-Opus-Deckard-Heretic-Uncensored-Thinking-NEO-CODE-Di-IMatrix-MAX-GGUF
hf.co/mradermacher/Llama3.3-8B-Instruct-Thinking-Heretic-Uncensored-Claude-4.5-Opus-High-Reasoning-i1-GGUF
```

I would suggest making it 128 instead of 80.

Other notes:

- `matches_filter` returning false on an invalid filter makes `llmman list <garbage>` print an empty table instead of an error.
- The scheme test comment says the gate keys on `:/` itself. There is no such gate; `hf:/owner/repo` is rejected by `check_host` on the empty port. Fix the comment.
- `ParsedRef` is `allow(dead_code)` outside tests and production only reads `.host`. Either thread the parsed type through the call sites or drop the unused fields.
- Producing 400 by downcasting `InvalidReference` inside `into_response` is indirect. Building the 400 where the reference is rejected reads better.
- Conflicts with #329 (now merged) in `serve.rs` on both unload paths. Needs a rebase.

The single-attempt metadata split is right and the part grammar is the correct shape.

PR has conflicts.

maxgio92:
@ericcurtin thanks for the top-500 check, all five points are fixed in b6ea871, rebased onto main.

1. MAX_PART_LEN is 128. Both refs you quoted now pass and are pinned in tests.
2. An invalid list filter now errors instead of printing an empty table.
3. Scheme comment rewritten; there is no ":/" gate.
4. ParsedRef is no longer dead_code: a debug_assert checks that rejoin() rebuilds every parsed input byte for byte, so production reads all fields. I can thread or drop the fields instead.
5. Each rejection site builds its 400; the into_response downcast is gone.

The rebase makes unload_key fallible, so both unload paths share the 400 mapping. The digest bound also rises to 128, still short of sha512; that question is still open.

ericcurtin:
Merged. Verified the merge with #323 builds and all 423 lib tests pass.

Follow-ons:
- Fuzz target for `parse_registry_ref`: no accepted parse yields a `..` or over-length part.
- Bound the OCI copy manifest phase to one attempt, same split as the metadata GET.
- Retry the metadata request once on a 401 auth challenge.
- Thread `ParsedRef` through the call sites so `has_host` stops reparsing per reference.
- `handle_pull`/`handle_push` can return `AppError` instead of hand-rolling the 400 JSON.
- Relative and Windows paths now hard-fail instead of silently becoming `hf.co/...`. Either support them or say so in the error.

PR #306, opened 2026-08-27
fix(serve): handle stream false on /api/chat and /api/generate

onuralpszr:
I tested with python ollama vs llmman  and request body;

`  {"model":"...","messages":[{"role":"user","content":"Say hi in three words."}],"stream":false}`

```
                          py-ollama (correct)   llmman before   llmman after
  /api/chat body          1 JSON object      12 NDJSON lines 1 JSON object
  /api/generate body      1 JSON object      12 NDJSON lines 1 JSON object
  response Content-Type   application/json   x-ndjson        application/json
  json.loads() on body    ok                 raises          ok
  streaming still works   ok                 ok              ok
```

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -6989,6 +6989,23 @@ mod tests {
     /// lines split across arbitrary TCP chunk boundaries must be
     /// reassembled, CRLF line endings trimmed, and a trailing
     /// unterminated line flushed when the stream ends.
+    /// A multi-byte character split across chunk boundaries must survive.
+    /// Decoding each chunk on its own turned the split halves into U+FFFD.
+    #[test]
+    fn bytes_to_lines_preserves_utf8_split_across_chunks() {
+        let line = "data: g\u{fc}nayd\u{131}n \u{1f600}";
+        // One byte per chunk: every multi-byte character is split.
+        let raw = format!("{line}\n");
+        let chunks: Vec<reqwest::Result<Bytes>> = raw
+            .as_bytes()
+            .iter()
+            .map(|b| Ok(Bytes::copy_from_slice(&[*b])))
+            .collect();
+        let stream = bytes_to_lines(futures::stream::iter(chunks));
+        let lines: Vec<String> = futures::executor::block_on(StreamExt::collect::<Vec<_>>(stream));
+        assert_eq!(lines, vec![line.to_string()]);
+    }
+
     #[test]
     fn bytes_to_lines_ported_ollama_client_stream_chunking() {
         let chunks: Vec<reqwest::Result<Bytes>> = vec![
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -1083,7 +1083,7 @@ fn image_data_uri(base64_bytes: &str) -> String {
     }
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Serialize, Default)]
 struct OAIChatRequest {
     model: String,
     messages: Vec<OAIMessage>,
@@ -6985,10 +6985,6 @@ mod tests {
         assert_eq!(ext.flush(), "");
     }
 
-    /// Ported from ollama's api/client_test.go (TestClientStream): SSE
-    /// lines split across arbitrary TCP chunk boundaries must be
-    /// reassembled, CRLF line endings trimmed, and a trailing
-    /// unterminated line flushed when the stream ends.
     /// A multi-byte character split across chunk boundaries must survive.
     /// Decoding each chunk on its own turned the split halves into U+FFFD.
     #[test]
@@ -7006,6 +7002,10 @@ mod tests {
         assert_eq!(lines, vec![line.to_string()]);
     }
 
+    /// Ported from ollama's api/client_test.go (TestClientStream): SSE
+    /// lines split across arbitrary TCP chunk boundaries must be
+    /// reassembled, CRLF line endings trimmed, and a trailing
+    /// unterminated line flushed when the stream ends.
     #[test]
     fn bytes_to_lines_ported_ollama_client_stream_chunking() {
         let chunks: Vec<reqwest::Result<Bytes>> = vec![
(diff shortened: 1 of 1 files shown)


ericcurtin (changes requested):
Reviewed at `0b18932`.

Good shape overall. Hoisting the per-line decode into `OllamaLineDecoder` so both paths read a response identically is better than a second reader that can drift, and `fold_ollama_lines` is pure and testable. I confirmed `stream` still defaults to `true` (serve.rs:709, :738), so omitted-`stream` clients keep streaming.

Two things to fix before merge.

## Blocking: truncated generations are returned as successful ones

`bytes_to_lines` collapses `Some(Err(_))` and `None` into one branch (serve.rs:2981), so a mid-generation transport failure looks like a clean EOF. Streaming survives this since the client never sees `done: true`. The non-streaming path can't, because it manufactures the terminal state: hardcoded `done: true`, `done_reason: "stop"`, HTTP 200. A connection reset halfway through and a complete answer are indistinguishable to the caller.

Contract to implement: when `stream` is false, abandon aggregation the instant the generation reports a failure and reply with a JSON error object carrying that status code, never a 200 with a partial answer. A non-streaming caller has no other way to find out the generation didn't finish.

Preferred fix: make `bytes_to_lines` yield `Result<String, _>` and have the non-streaming branch bail on the first `Err` into an `AppError`. That touches four call sites (:2939, :3322, :3332, :3395), but the other three can `filter_map` errors away to keep today's behavior. Driving `resp.bytes_stream()` directly in the non-streaming branch also works, but re-introduces the two-readers problem this PR removes.

## Content dropped when a response ends without `done`

`RawContentExtractor` buffers content for tag disambiguation and is only drained on `done`. `fold_ollama_lines` never flushes after the loop. Verified on the branch:

```rust
let lines = [
    r#"data: {"choices":[{"delta":{"content":"<"},"finish_reason":null}]}"#,
    r#"data: {"choices":[{"delta":{"content":"thi"},"finish_reason":null}]}"#,
].map(String::from);
// content == ""   ("<thi" is a strict prefix of "<think>", never flushed)
```

Flush unconditionally after the loop. `flush` drains via `mem::take`, so it's already idempotent. Worth a test.

## Smaller

* serve.rs:3042: the `post_chat` doc comment still lists `collect_completion`, which this PR deletes.
* serve.rs:3323: `collected: Vec<String>` buffers the whole response for no reason. Goes away with the fix above.
* serve.rs:3321: `let _activity = activity;` is correct but reads like an unused-variable silencer. Its counterpart at :3435 has a comment.

## Coverage

The fold is well covered, the wiring isn't. Nothing checks that `stream: false` returns one JSON object with `application/json` rather than NDJSON, `/api/generate`'s non-streaming path is untested (different `build_chunk` closure), and nothing pins `stream` defaulting to `true`. That last one is newly load-bearing in this PR and is a two-line `from_str::<OllamaChatRequest>` assertion.

Happy to re-review once error propagation is in.

ericcurtin (approved):
Re-reviewed at `f464fd3`. Blocking issue is resolved. Approving.

You solved the truncation problem differently from what I suggested and I think your way is better. Rather than making `bytes_to_lines` fallible and touching four call sites, you treat the absence of a terminal chunk as the evidence of truncation and 502 on it. Same guarantee, no signature churn, and the comment explaining why a missing `done` is the only available signal is exactly the right thing to leave behind.

Verified:

* A backend dying mid-response leaves `fold.done` false and returns 502 instead of a 200 carrying half an answer. `fold_ollama_lines_reports_a_missing_terminal_chunk` pins it.
* The dropped-content bug I raised is resolved as a consequence. Content buffered in `RawContentExtractor` was only unreachable on the no-`done` path, and that path now errors instead of returning a silently short answer.
* `OllamaFold` accumulating `done` with `|=` is right given `decode` can report done more than once, and the tool-call drain still survives the trailing `[DONE]`.
* `bytes_to_lines_preserves_utf8_split_across_chunks` is a good test. One byte per chunk is the right way to write it, and it genuinely fails against the old per-chunk `from_utf8_lossy`.
* Full suite green locally (339 passed), CI green on all five targets. The one clippy warning I saw is pre-existing in `webui.rs`, unrelated.

Merging. Follow-ups in a separate comment, none of them blocking.

ericcurtin:
Merged as 6e104c5.

First, a correction on my part. I reviewed and approved against `f464fd3`, but you pushed `7057d30` in between and the merge picked that up, so I approved a commit I had not actually read. My mistake, I should have re-checked the head SHA immediately before merging. I have reviewed it since and it is good: the two `stream_ollama` tests over a real mock backend are exactly the HTTP level coverage that was missing, and moving the UTF-8 test below `bytes_to_lines_ported_ollama_client_stream_chunking` fixed a doc comment that would otherwise have ended up describing the wrong test. 341 tests pass on merged main. No action needed, flagging it only because approving unread code is not a thing I want to do silently.

Remaining follow-ups, none blocking, roughly in priority order.

**1. `stream` defaulting to `true` is still untested and newly load-bearing.**
Before this PR `stream_ollama` always streamed and `req.stream` was inert, so `#[serde(default = "bool_true")]` (serve.rs:709, :738) had no observable effect. It does now. If that attribute is ever dropped or weakened to a plain `#[serde(default)]`, every client that omits `stream` silently switches to non-streaming and the suite stays green. Two line `serde_json::from_str::<OllamaChatRequest>` assertion per struct. Highest value gap left.

**2. The handler to `stream_ollama` wiring is still uncovered.**
`run_stream_ollama` calls `stream_ollama` directly, so its doc comment claiming the tests would catch "a handler's flag forwarding" regressing is a slight overclaim: nothing exercises `handle_ollama_chat` or `handle_ollama_generate` passing `req.stream` through. Related, `/api/generate` is still untested on this path, and it supplies a different `build_chunk` closure to a different response shape, so the chat test passing does not imply generate works.

**3. The 502 assumes the backend terminates its SSE stream.**
Truncation is inferred from a missing terminal chunk, correct for llama-server since it always closes with `data: [DONE]`. A backend emitting neither `[DONE]` nor a `finish_reason` would 502 every non-streaming request despite the content arriving intact. Low risk as both are standard OpenAI SSE, but worth knowing if the MLX or vllm-plugin backends get exercised through `stream: false`.

**4. `#[derive(Default)]` on `OAIChatRequest`.**
Added for test ergonomics but it widens a production type: `OAIChatRequest::default()` is now silently constructible with `model: ""` and `stream: false`, neither of which is ever valid to send. Consider a test-only builder instead, or `#[cfg_attr(test, derive(Default))]`.

**5. One doc comment still documents the wrong test.**
serve.rs:6613, "Ported from ollama's api/client_test.go ... a bad chunk must never abort the whole stream" now sits on `fold_ollama_lines_concatenates_deltas` but describes `oai_chunk_to_content_ported_ollama_stream_decoding_cases`, which lost its comment when the fold tests were inserted above it. Same insertion pattern you already fixed for the UTF-8 test.

**6. Stale reference to a deleted function.**
serve.rs:3042 still lists `collect_completion` as one of `post_chat`'s callers. This PR deleted it.

**7. `done_reason` is always `"stop"` or `"tool_calls"`.**
`build_chunk` synthesizes it rather than carrying through the upstream `finish_reason`, so a response cut short by `num_predict` reports `"length"` as `"stop"`. Pre-existing and shared with the streaming path, not introduced here, but more visible now that the non-streaming reply is a single object where that field is the only completion signal.

**8. Minor.**
`let _activity = activity;` at serve.rs:3337 is correct and holds the guard past the awaits, but bare like that it reads as an unused-variable silencer and invites deletion. Its counterpart at :3470 has a comment explaining exactly that.

PR #317, opened 2026-08-27
fix(daemon): signal the daemon's process group via kill(2), not kill(1)

ericcurtin:
Every Linux E2E leg on `main` has failed since #291 with no test failure of its own — just `The runner has received a shutdown signal`. **llmman was killing the GitHub Actions runner.**

## Root cause

`signal_group`/`kill_daemon` shelled out to `kill -TERM -<pid>`. procps-ng's `kill(1)` reads only the **first digit** of a negative pid after a `-SIG` flag:

```console
$ strace -e trace=kill kill -TERM -12345
kill(-1, SIGTERM)          # "signal every process this user owns"
```

CI daemon pids all led with `1`, so dead-daemon cleanup became a user-wide SIGTERM, then a user-wide SIGKILL when `stop_group` escalated. BSD `kill(1)` parses the same argv correctly, hence Linux-only. Any Linux user whose `llmman serve` died at startup hit it too.

## Fix

`kill(2)`: no argv to misparse, and a truthful `ESRCH` where procps-ng exits `0` even for a nonexistent group — so `stop_group`'s escalation check never worked either. `kill_group` refuses pids `0` and `1`, capping the blast radius by construction.

## Verification

E2E only runs on push-to-main, so PR CI cannot exercise this. Verified locally against the real E2E test in a PID namespace with the pid counter set to a CI-like `12345`:

```
WITHOUT FIX: test ... Terminated       RUNNER-STANDIN: KILLED
WITH FIX:    test ... ok (0.6s)        RUNNER-STANDIN: SURVIVED
```

Three new tests cover the refused pids, an empty group reporting unsignalled, and a group kill that spares a bystander.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -265,18 +265,17 @@ fn wait_for_port_free() -> bool {
 
 #[cfg(unix)]
 fn kill_daemon(identity: &DaemonIdentity, force: bool) {
-    let signal = if force { "-KILL" } else { "-TERM" };
+    let signal = if force { libc::SIGKILL } else { libc::SIGTERM };
     match identity.pid {
-        // Negative pid = the whole process group (the daemon is its own
-        // group leader), falling back to just the pid if that fails.
+        // The daemon's whole process group, falling back to the bare pid
+        // — which /api/version just proved alive — if it leads no group.
         Some(pid) => {
-            let group = Command::new("kill")
-                .args([signal, &format!("-{pid}")])
-                .status();
-            if !group.map(|s| s.success()).unwrap_or(false) {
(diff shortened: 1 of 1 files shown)


ericcurtin:
Declining the suggestion to drop `kill_daemon`'s single-process fallback: that pid comes from a daemon that just answered `/api/version`, so it is a live process, not a reaped pid awaiting reuse. The fallback is what handles a daemon that isn't leading its own group. `signal_group`, whose callers *can* hold a reaped pid, still has no fallback.

Shortened the commit message, description and code comments.

PR #318, opened 2026-08-27
feat(launch): add --provider for models.dev hosted providers

ericcurtin:
## What

`llmman launch` could only point an integration at a model llmman serves itself. `--provider` extends it to hosted models, resolved from the same [models.dev](https://models.dev) catalog `opencode` uses:

```
export OPENROUTER_API_KEY=...
llmman launch opencode --provider openrouter --model qwen/qwen3-coder
```

`llmman launch --list-providers` prints what's available, with each provider's API-key variable and whether it's set.

## Design decisions

**The catalog is fetched at runtime, not hardcoded.** opencode doesn't hardcode its providers either — it GETs `api.json` and caches it. Same here, so a provider models.dev gains tomorrow works without an llmman release. Cached 24h (written through a temp file and a rename, since `launch` and `serve` refresh it independently), with a stale copy used if the fetch fails, so being offline degrades to an out-of-date list rather than a broken command. There is deliberately no URL override: the catalog has one source, and an override is a way to point llmman at a list that disagrees with the rules its routing was written against.

**Requests still go through `llmman serve`.** `--provider` changes where the daemon forwards them, not who the integration talks to. One endpoint and one place integrations are configured whether a model is local or hosted, and both usable from the same session.

**Only providers llmman can be sure it reaches are offered.** The daemon speaks one thing upstream: an OpenAI Chat Completions POST with `Authorization: Bearer`. An entry qualifies only if its `npm` driver is OpenAI-wire-format, it has one concrete `https` base URL, and one API-key variable — **175 of today's 207**. The rest (Bedrock's SigV4, Vertex's GCP credentials, Azure's per-account endpoints, the `@ai-sdk/anthropic` gateways, GitHub Copilot's OAuth token exchange, and the loopback entries llmman serves itself anyway) are absent rather than half-supported. A small `BUILTIN_ENDPOINTS` table supplies the endpoint for the 13 majors models.dev leaves `api` unset for; it is not a second registry, since an entry still has to be in the fetched catalog to be offered.

**Provider-routed references are namespaced under `llmman.provider/`.** The bare `<provider>/<model>` spelling opencode uses is ambiguous here: llmman also resolves two-segment names as HuggingFace repos, so routing on a bare prefix would silently capture `google/gemma-3`, `openai/gpt-oss-120b` and every other real `hf.co/<org>/<repo>` whose org shares a provider's name. Covered by a regression test.

**The API key travels per request**, in the integration's own `Authorization`/`x-api-key` header, falling back to the daemon's environment. That's what makes this work against an *already-running* daemon — the common case, since `ensure_server` reuses a live one — and means the key is never persisted. The `llmman` placeholder locally-served integrations get is explicitly not a credential, so it's never forwarded to a real provider.

## Implementation

- **`src/providers.rs`** (new) — catalog fetch, cache, and the reachability filter, plus the reference encoding.
- **`src/cmd/serve.rs`** — what was a bare `u16` loopback port is now a `Target`: that same local backend, or a remote provider. The four functions that actually send a request build their URL and attach credentials through it. `/api/show` answers for provider-routed refs so `ensure_model_pulled` doesn't try to pull one, and preload skips them.
- **`src/cmd/launch.rs`** — the flag, the resolved key handed to the integration, and validation that fails in llmman's own output (naming the missing variable, suggesting near-miss provider ids, listing example models) before the integration takes over the terminal.

## Behavior without `--provider` is unchanged

`Target::Local` produces byte-for-byte the same loopback URLs, models resolve exactly as before, no credential is attached, a local backend's failures keep the status they always returned, and the llama-server request fixups (`repeat_penalty`, `/v1/responses` tool filtering) still apply exactly as before. Covered by tests.

## Testing

- Lib tests, `cargo fmt --check`, and `--features podman --no-default-features` all clean; `launch_e2e` still compiles.
- **54 new unit tests.** The filter rules are pinned one fixture each, then run together against a real trimmed `api.json` (`tests/fixtures/`) — that last one asserts the majors come out at the right endpoints, the unreachable categories stay out, and every offered provider has an https, non-templated, non-trailing-slash URL and a key variable. Also URL re-basing for `/v1`, `/v1beta/openai` and bare-host providers, header key extraction and its placeholder/non-bearer guards, prefix round-tripping, the HuggingFace-collision regression, and a real HTTP round trip asserting what a provider actually receives.
- **Manually verified** every `BUILTIN_ENDPOINTS` URL answers `POST /chat/completions` with a bearer token (13/13 return 401/400, i.e. the route exists and parsed the request), and each error path (`--list-providers`, unknown provider, missing `--model`, missing key, unlisted model) in llmman's own output.

I did not add a `launch_e2e` case, since that suite drives real integration binaries against real models — happy to add one if you'd like it there.

## Since review

Two rounds. The first addressed the CodeRabbit/Copilot comments plus a self-review against opencode's own `models-dev.ts`; the second addressed Copilot's re-review.

**Round 1**
- Provider endpoints must be `https` — the catalog is fetched at runtime, so an `http://` entry upstream would have been enough to put a key on the wire in the clear.
- The catalog is parsed per entry, so one malformed upstream record costs that provider rather than the whole list.
- No longer memoized for the life of the process: a success is re-checked after 24h and a failure retried after 60s.
- Cache writes go through a temp file and a rename, as opencode's do — `launch` and `serve` refresh it independently and a plain write leaves a torn-read window.
- **Dropped the `v0` builtin**: nothing answers at its documented `https://api.v0.dev/v1` (404, where the other 13 return 401/400), so llmman had no endpoint it could vouch for.
- `repeat_penalty` (a llama.cpp extension) is dropped for remote requests — OpenAI rejects unrecognized arguments outright.
- A provider's own 4xx passes through instead of being buried in llmman's blanket 500, and errors name the provider rather than "llama-server". A bad key reads as 401, not "internal server error".
- The resolved provider key is handed to the integration rather than the `llmman` placeholder, so `--provider` works against an already-running daemon. Integrations configured through a file on disk keep the placeholder rather than persist a credential.
- `/v1/audio/transcriptions` refuses a provider-routed model instead of forwarding a multipart body whose `model` field the provider has never heard of.
- `RemoteTarget`'s `Debug` redacts the key; `client_api_key` filters each header before choosing between them; the URL re-basing rule lives in one place.
- Corrected a factual claim: opencode fetches `models.opencode.ai/api.json`, its own mirror — byte-identical to `models.dev`'s today (I diffed them), but not "the same source".

**Round 2**
- `load()` fell back to a stale cache and reported plain success, so a transient outage was cached for 24h rather than retried after 60s. It now distinguishes `Fresh` from `Stale`, and stale gets a failure's short lifetime.
- Fetch and parse are one refresh step: a 200 carrying a truncated body or a captive portal's HTML now reaches the stale cache instead of failing outright.
- `/v1/responses` sanitization is local-only. `filter_non_function_tools`/`consolidate_responses_instructions` work around what *llama-server* rejects, and ran before the target was known — a provider with native Responses support silently lost `web_search` and every other non-function tool.
- The daemon no longer spends its own key for callers it cannot authenticate: with a non-loopback bind, the environment fallback is not consulted and the 401 says so.
- `--provider` requires a loopback daemon, since with a remote `LLMMAN_HOST` the key would cross the network in cleartext. Two different questions, so two predicates — `connects_over_loopback` (is *our* hop local; a wildcard bind counts) gates launch, `reachable_only_locally` (can anyone *else* reach this daemon; a wildcard bind does not) gates the server-side fallback — with a test on the distinction.
- `--provider` refuses `cline`, `kimi` and `copilot` rather than launching a configuration that cannot work: the first two select their own model, and `copilot` cannot carry a key.

**Round 3**
- `bailing`'s `api` is a full route rather than a base URL (`.../v1/chat/completions`), so every request to it would have gone to `.../chat/completions/chat/completions`. It is the only such entry in the live catalog today; `base_url_of` normalizes the suffix away, and a test walks the whole real fixture asserting no provider can build a doubled route.
- `Bearer` is matched case-insensitively (RFC 7235) — a client sending `bearer` had its key silently dropped, with no fallback on a daemon that won't use its own.
- `--provider` refuses `hermes`/`openclaw` on a wildcard bind: they are configured through a file so they send the placeholder, and the daemon won't spend its own key for a caller anyone could be. That combination used to 401 from inside the integration.
- The three proxy helpers name the actual target instead of blaming `llama-server` for a provider failure.
- The locality predicates delegate to host-taking forms so the test exercises them directly, rather than the private pieces they were built from.
- README fence labelled `sh` (MD040).

**Round 4**
- An empty catalog after filtering counted as a successful refresh. `{"error":"temporarily unavailable"}` is valid JSON and a valid map, so llmman would have overwritten a good cache with nothing and memoized it for 24h. `from_json` now errors when nothing is routable, so the stale copy survives.
- `github-copilot` excluded: it advertises `@ai-sdk/openai-compatible` + `GITHUB_TOKEN`, but that bearer 401s against `api.githubcopilot.com` — opencode trades it through `login/oauth/access_token` first. Providers that only add optional attribution headers (`kilo`, `openrouter`, ...) are unaffected.
- Cross-site requests no longer reach the daemon-environment key: CORS gates reading a reply, not sending a request, and these handlers never check `Content-Type`, so a page the user merely visits could have spent their provider credits. Gated on `Sec-Fetch-Site`, which browsers always send and CLI integrations never do.
- `gemini` refused for `--provider`: its key feeds a native Google client, and being wrong about `GEMINI_BASE_URL` would send a third-party key to Google. The placeholder was harmless either way; a real key is not.
- Declined the Windows `fs::rename` report — `std::fs::rename` is documented to replace an existing destination, via `MoveFileExW`/`SetFileInformationByHandle`.

**Round 5**
- `openclaw` refused for `--provider`: `launch_openclaw` passes `--custom-model-id` only through first-run onboarding, so every later launch would quietly reuse the old model instead of the provider reference.
- The Responses-API gap turned out to be concrete, and I was wrong to wave it off in round 1. Probing it: `anthropic` and `mistral` 404 on `/v1/responses`, `openai`/`groq`/`openrouter` answer it — so `llmman launch codex --provider anthropic` failed on its first request with a bare 404. Rather than hardcode a capability list models.dev cannot back (and that would be wrong the week a provider ships the route), llmman now reports the 404 it actually received as a 501 naming the provider and what it is missing.

**Round 6**
- `explain_missing_route` ignored its `route` argument (a bug from round 5), so any provider 404 — an unknown model on `/v1/chat/completions`, say — would have been rewritten as "lacks the Responses API". Now gated on the route, with tests covering the other routes and statuses passing through untouched.
- `/v1/responses/input_tokens` goes through the other proxy helper and bypassed that handling entirely; both are wired up now.
- The `hermes` key check was the wrong check: requiring the key in the launching shell rejected a good daemon that has it, while having it there guaranteed nothing about the daemon. For integrations that cannot carry a key it is now a warning naming the daemon requirement; ones that can still fail fast.
- Stopped implying the loopback check is authentication. It bounds the blast radius — network-reachable binds and cross-site browser requests get nothing — but any local account on a shared machine can already reach a loopback daemon for everything else it does. Comment and README now say so and point at per-request keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,61 @@ Run `llmman launch` with no arguments to list the supported integrations (Claude
 
 Short names work wherever a model reference is accepted.
 
+#### Hosted providers
+
+`--provider` points the same integrations at a model llmman doesn't serve
+itself:
+
+```sh
+export OPENROUTER_API_KEY=...
+llmman launch opencode --provider openrouter --model qwen/qwen3-coder
+```
+
+The provider list is fetched at runtime from
+[models.dev](https://models.dev) — the same catalog `opencode` resolves
+its own providers from — so a newly added provider works without an
+llmman release. It's cached for 24 hours, and a stale copy is used if the
+fetch fails, so being offline means an out-of-date list rather than a
+broken command. `llmman launch --list-providers` prints the providers
+llmman can route to, with each one's API-key variable and whether it's
+set.
+
+Requests still go through `llmman serve`; `--provider` changes where the
+daemon forwards them, not who the integration talks to. So one endpoint
+and one place integrations are configured, whether a model is local or
+hosted, and both usable from the same session.
+
+The API key is read from the variable models.dev names for that provider
+and travels per request, never to disk. `hermes` is the exception: llmman
+configures it through a file on disk, so it can't carry a key and
+`llmman serve` needs the variable in its own environment instead. That
+fallback is only used for a daemon bound to loopback, and never for a
+browser request from another site — it bounds the blast radius rather
+than authenticating anyone, so on a shared machine prefer an integration
+that sends its own key. `cline`, `kimi`, `copilot`, `gemini` and `openclaw` can't be
+used with `--provider` at all — the first two pick their own model rather
+than taking llmman's, `copilot` has no way to send a key, `gemini` feeds
+its key to a native Google client llmman can't confirm it has redirected,
+and `openclaw` only takes a model during first-run onboarding.
+
+Being OpenAI-compatible doesn't mean implementing every OpenAI route.
+`codex` uses `/v1/responses`, which `openai`, `groq` and `openrouter`
+answer but `anthropic` and `mistral` don't; models.dev carries no
+capability data to filter on, so llmman turns that provider's bare 404
+into an explanation naming what's missing rather than guessing up front.
+
+`--provider` needs a local `llmman serve`. The daemon talks plain HTTP
+and has no authentication, so llmman will not hand an integration a real
+key to send to a remote `LLMMAN_HOST`, and a daemon bound to anything but
+loopback will not spend its own environment's key on behalf of a caller
+that didn't present one.
+
+Providers llmman cannot reach with a single bearer token over an
+OpenAI-compatible https endpoint are deliberately absent rather than
+half-supported: Amazon Bedrock (SigV4 request signing), Google Vertex
+(GCP service-account credentials), Azure, and the others whose endpoint
+is per-account or whose wire format isn't OpenAI's.
+
 ### Use with vLLM directly
 
 `llmman serve` already spawns `vllm` itself as a backend for safetensors
--- a/src/cmd/launch.rs
+++ b/src/cmd/launch.rs
@@ -2,6 +2,14 @@
 //!
 //! Mirrors `ollama launch`: sets integration-specific environment variables
 //! pointing at the local inference server, then exec's the integration binary.
+//!
+//! `--provider` extends that to models llmman does not serve itself, from
+//! the same models.dev catalog opencode resolves its providers from (see
+//! [`crate::providers`]). It does not change the shape above: the
+//! integration is still pointed at `llmman serve`, which forwards upstream
+//! on its behalf. There is deliberately no path here that hands an
+//! integration a provider's URL directly — one endpoint, one place
+//! integrations are configured, whether or not the weights are local.
 
 use std::path::PathBuf;
 use std::process::Command;
@@ -10,6 +18,7 @@ use anyhow::Context;
 use clap::Args;
 
 use crate::daemon;
+use crate::providers::{self, Provider};
 
 // ---------------------------------------------------------------------------
 // CLI
@@ -26,31 +35,68 @@ pub struct LaunchArgs {
     #[arg(long, short, value_name = "MODEL")]
     pub model: Option<String>,
 
+    /// Serve --model from this provider (openai, anthropic, openrouter, …)
+    /// instead of locally. Requires --model. See --list-providers.
+    #[arg(long, short = 'p', value_name = "PROVIDER")]
+    pub provider: Option<String>,
+
+    /// List the providers --provider accepts, and exit
+    #[arg(long)]
+    pub list_providers: bool,
+
     /// Extra arguments forwarded to the integration binary (after --)
     #[arg(last = true, value_name = "ARGS")]
     pub extra_args: Vec<String>,
 }
 
 pub fn run(args: &LaunchArgs) -> anyhow::Result<()> {
+    if args.list_providers {
+        return print_providers();
+    }
+
+    let provider = args
+        .provider
+        .as_deref()
+        .map(str::trim)
+        .filter(|p| !p.is_empty());
+
     let Some(ref name) = args.integration else {
         print_integrations();
         return Ok(());
     };
 
-    // resolve_ollama_api, not resolve: every integration this launches talks
-    // to serve's Ollama/OpenAI/Anthropic-compat surfaces, all of which
-    // resolve model names the same way (see ensure_model in cmd::serve), so
-    // a bare name here must match what the daemon resolves it to at
-    // request time.
-    let model = args
-        .model
-        .as_deref()
-        .map(crate::shortnames::resolve_ollama_api)
-        .unwrap_or_default();
+    let (model, api_key) = match provider {
+        Some(provider) => {
+            check_provider_supported(name)?;
+            let per_request = !PROVIDER_NEEDS_DAEMON_KEY.contains(&name.to_lowercase().as_str());
+            resolve_provider_model(provider, args.model.as_deref(), name, per_request)?
+        }
+        // resolve_ollama_api, not resolve: every integration this launches
+        // talks to serve's Ollama/OpenAI/Anthropic-compat surfaces, all of
+        // which resolve model names the same way (see ensure_model in
+        // cmd::serve), so a bare name here must match what the daemon
+        // resolves it to at request time.
+        None => (
+            args.model
+                .as_deref()
+                .map(crate::shortnames::resolve_ollama_api)
+                .unwrap_or_default(),
+            providers::PLACEHOLDER_API_KEY.to_string(),
+        ),
+    };
 
     // Ensure serve is running (start it in background if needed), preloading
     // the requested model so the integration's first request finds it warm.
-    crate::daemon::ensure_server(&model)?;
+    //
+    // A provider-routed model is never preloaded: there is nothing local to
+    // warm up, and asking the daemon to load one would just fail. The daemon
+    // still has to be running — it is what forwards upstream.
+    let preload = if provider.is_some() {
+        ""
+    } else {
+        model.as_str()
+    };
+    crate::daemon::ensure_server(preload)?;
 
     // serve's preload above is fire-and-forget and only fires on a cold
     // `serve` start (see run() in cmd/serve.rs) — if the daemon was already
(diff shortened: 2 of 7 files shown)


ericcurtin:
Addressed: https-only provider endpoints, per-entry catalog parsing, catalog refresh/retry instead of a process-lifetime `OnceLock`, `repeat_penalty` dropped for remote requests, the resolved provider key handed to integrations rather than the `llmman` placeholder, transcriptions refused for provider-routed models, redacted `RemoteTarget` `Debug`, per-header key filtering, and one shared URL re-basing rule. Also rebased on main and squashed.

Declining two:

- **Restrict provider/integration combinations by endpoint support** — models.dev carries no endpoint capabilities, so any list of "providers that implement `/v1/responses`" would be hardcoded guesswork that goes stale silently. A provider that does not implement the route answers 404, which is a clearer signal than llmman refusing a combination that may well work.
- **Rewrite `response.model` in Responses API SSE** — the same gap already exists for `Engine::Mlx`, which is the pre-existing user of `response_model_override`. It is cosmetic (the client sees the upstream id in one field), unrelated to provider routing, and belongs in its own change.

Shortened the commit message, description and code comments.

ericcurtin:
@coderabbitai review

ericcurtin:
Self-reviewed against `opencode`'s own `packages/core/src/models-dev.ts`, which turned up three things beyond the review comments:

- **Dropped the `v0` builtin.** Probing all 14 builtin endpoints with a bearer token: 13 answer `POST /chat/completions` with 401/400 (route exists, request parsed, key rejected), `https://api.v0.dev/v1` returns 404 for `/chat/completions`, `/models`, and the un-versioned path. An endpoint llmman cannot vouch for is exactly what the "vetted by hand" rule exists to keep out, so `v0` is now absent like the other unreachable entries.
- **Atomic cache writes.** opencode writes its copy through a temp file and a rename (plus a flock). `llmman launch` and `llmman serve` refresh the catalog independently and do overlap, so the plain `fs::write` left a window for a torn read. Now temp + rename.
- **Corrected a factual claim.** opencode fetches `models.opencode.ai/api.json`, its own mirror — not `models.dev/api.json`. Byte-identical today (I diffed them), but the comment claiming "the same source" was wrong, and llmman reads upstream directly on purpose.

Also added a trimmed real `api.json` fixture and a test that runs the whole filter against it, rather than only hand-written per-rule fixtures — verified non-vacuous by removing the https rule and watching it fail. And a provider's own 4xx now passes through instead of being buried in llmman's blanket 500, so a bad key reads as 401 rather than "internal server error".

Shortened the commit message, description and code comments.

ericcurtin:
Thanks — six good catches, all addressed.

**`load()` fell back to a stale cache and reported success**, so a transient outage was cached for 24h rather than retried after 60s. `load` now returns `Loaded::Fresh` or `Loaded::Stale`, and `catalog()` gives stale the same short lifetime as a failure. Pinned by a test, since collapsing the two back is the easy mistake.

**Fetch and parse are now one refresh step.** A 200 carrying a truncated body or a captive portal's HTML failed straight out instead of reaching the stale cache. Both are now the same failure path.

**`/v1/responses` sanitization is local-only.** `filter_non_function_tools`/`consolidate_responses_instructions` are workarounds for what *llama-server* rejects, and they ran before the target was known — a provider with native Responses support would silently lose `web_search` and every other non-function tool. Both now run after target resolution, on local targets only.

**The daemon no longer spends its own key for callers it cannot authenticate.** With a non-loopback bind, `provider.api_key()` is not consulted, and the 401 says so; such a caller has to present its own.

**`--provider` requires a loopback daemon.** With a remote `LLMMAN_HOST` the key would cross the network in cleartext before llmman ever saw it. Note these are two different questions, so two predicates: `connects_over_loopback` (is *our* hop local — a wildcard bind counts) gates launch, `reachable_only_locally` (can anyone *else* reach this daemon — a wildcard bind does not count) gates the server-side fallback. Collapsing them would either leak credentials or refuse a perfectly local hop, so there is a test on the distinction.

**`--provider` now refuses `cline`, `kimi` and `copilot`** rather than launching a configuration that cannot work — the first two select their own model so the namespaced reference never reaches the daemon, and `copilot` has no way to carry a key. Same call the catalog filter makes: absent beats half-supported. A test asserts each refused name is one `launch` actually dispatches.

Shortened the commit message, description and code comments.

ericcurtin:
All six addressed.

**`bailing`'s `api` is a full route, not a base URL** (`.../v1/chat/completions`), so every request to it would have gone to `.../chat/completions/chat/completions`. Confirmed against the live catalog — it is the only such entry today. `base_url_of` now normalizes the suffix away, with a regression test plus one that walks the whole real fixture asserting no provider can produce a doubled route.

**`Bearer` is matched case-insensitively** — RFC 7235 makes the scheme case-insensitive, and on a daemon that won't use its own key there was nothing to fall back to. Tested across spellings, and the non-bearer test now also covers a bare `Bearer` and a scheme-less value.

**hermes/openclaw with a wildcard bind** would have 401'd: they are configured through a file so they send the placeholder, and the daemon refuses its own key on a bind anyone can reach. `--provider` now refuses that combination up front and names both halves of the reason, rather than letting it surface from inside the integration.

**The three proxy helpers name the actual target** — I had fixed `post_chat` but not these, so a provider connection failure still read as a llama-server problem.

**The locality predicates are now directly tested.** They delegate to host-taking forms, so the test exercises `host_connects_over_loopback`/`host_reachable_only_locally` for every case instead of the private pieces they were built from — including the wildcard host, which is the only one they disagree on and the whole reason both exist.

**README fence** labelled `sh` for MD040.

Shortened the commit message, description and code comments.

ericcurtin:
Four of five addressed; declining the Windows one with evidence.

**Empty catalog after filtering was treated as a successful refresh.** Correct and the worst of the five: `{"error":"temporarily unavailable"}` is valid JSON and a valid map, every entry drops out, and llmman would overwrite a good cache with nothing and memoize it for 24h. `from_json` now errors when nothing is routable, so `load` keeps the stale copy. Tested against that body, a rate-limit body, and `{}`.

**`github-copilot` is excluded.** Confirmed: it advertises `@ai-sdk/openai-compatible` + `GITHUB_TOKEN`, but `POST https://api.githubcopilot.com/chat/completions` with a `GITHUB_TOKEN` bearer returns 401 — opencode trades it through `login/oauth/access_token` first (`packages/opencode/src/plugin/github-copilot/copilot.ts`) and adds `X-GitHub-Api-Version`/`x-initiator`/`Openai-Intent`. Exporting the documented variable and getting an unexplained 401 is exactly what the filter exists to prevent. Note the attribution-header providers (`kilo`, `openrouter`, `zenmux`, ...) are *not* affected — those headers are optional, auth is still a plain bearer.

**Cross-site requests no longer reach the daemon-environment key.** Right that `CorsLayer` gates reading the response, not sending the request — and these handlers never check `Content-Type`, so a simple POST skips preflight entirely. Gated on `Sec-Fetch-Site` rather than `Origin`: browsers always send it, `same-origin`/`none` for llmman's own web UI and the address bar, `cross-site`/`same-site` for another page. CLI integrations send it never, so nothing legitimate loses the fallback, and a caller with its own key is unaffected either way.

**`gemini` refused for `--provider`.** I could not verify the CLI still honours `GEMINI_BASE_URL`, and the asymmetry decides it: being wrong means someone's OpenRouter key goes to Google. The placeholder this used to pass was harmless whichever endpoint it reached; a real key is not.

**Declining the Windows `fs::rename` claim** — it is incorrect. `std::fs::rename` is documented as "Renames a file or directory to a new name, **replacing the original file if `to` already exists**", implemented with `MoveFileExW`/`SetFileInformationByHandle`; the only documented Windows caveat is that `to` must not be a directory. Refreshes are not broken there.

Shortened the commit message, description and code comments.

ericcurtin:
Two new ones here; the rest of that batch is GitHub re-anchoring already-fixed comments onto the new commit (repeat_penalty, transcriptions, https-only, per-entry parsing, URL dedup, empty catalog, github-copilot, gemini, key propagation, Windows rename — all addressed in earlier rounds).

**`openclaw` is now refused for `--provider`.** Correct: `launch_openclaw` passes `--custom-model-id` only through onboarding, which runs once. Every later launch reuses whatever `openclaw.json` already names, so the provider reference would never reach the daemon and the session would quietly run on the old model — worse than failing. Moved out of the "needs the daemon key" list into the refused list.

**The Responses-API gap is real, and I was wrong to wave it off earlier.** I probed it: `anthropic` and `mistral` 404 on `/v1/responses`, while `openai`, `groq` and `openrouter` answer it. Since `codex` uses only that route, `llmman launch codex --provider anthropic` fails on its first request. My previous reasoning — "a 404 is a clear enough signal" — does not survive seeing it: the user gets a bare 404 with no indication which of the two things is missing.

I still do not want a hardcoded capability list (models.dev has no such data, and it would be wrong the week a provider ships the endpoint). So llmman reports the 404 it actually received instead of predicting it: a remote 404 on `/v1/responses` becomes a 501 naming the provider, saying it is OpenAI-compatible but does not implement the Responses API, and pointing at ones that do. Self-maintaining, and it fails in llmman's own words like the rest of this feature.

Shortened the commit message, description and code comments.

ericcurtin:
All four addressed — the first is a bug I introduced last round.

**`explain_missing_route` ignored its `route` argument.** Exactly right, and my mistake: the guard checked only status and remoteness, so a provider 404 for an unknown model on `/v1/chat/completions` would have been rewritten to "this provider lacks the Responses API" — a worse answer than the provider's own. Now gated on `is_responses_route`, with the test covering `/v1/chat/completions`, `/v1/embeddings`, a local 404, and non-404 statuses all passing through untouched.

**`/v1/responses/input_tokens` bypassed it** — it goes through `proxy_openai_passthrough`, which I had not wired up. Both helpers now post-process, and the test covers the token-counting route.

**The `hermes` key check was the wrong check.** Correct on both halves: requiring the key in this shell rejected a good daemon that has it, and having it here guaranteed nothing about the daemon. For integrations that cannot carry a key, a missing local key is now a warning naming the daemon requirement rather than an error — the daemon's own 401 already names the variable. Integrations that do carry one still fail fast.

**README contradiction** fixed: only `hermes` is listed as relying on the daemon's key; `openclaw` is refused outright.

On **loopback reachability not being authentication** — agreed, and I have stopped implying otherwise. The comment now says what is actually true: withholding the key from a network-reachable bind and from cross-site browser requests bounds the blast radius, it does not authenticate anyone, and on a shared machine any local account can already reach a loopback daemon for everything else it does. The README says the same and points at sending a key per request instead. Adding real authentication to `llmman serve` is a bigger change than this PR, and pretending otherwise in a comment would be worse than stating the limit.

Shortened the commit message, description and code comments.

ericcurtin:
The 18 comments now showing against the current commit are GitHub re-anchoring unresolved threads onto the new SHA — no new review has been submitted on it. Verified each one against the code as it stands:

| Finding | Where it is fixed |
|---|---|
| plaintext provider endpoints | `api.starts_with("https://")` in `routable` |
| whole-catalog parse failure | per-entry `from_value` in `from_json` |
| duplicated URL rule | `providers::rebase_url`, one definition |
| `repeat_penalty` sent to providers | `repeat_penalty_applies` |
| transcriptions forwarding a prefixed model | refused with a named error |
| `/v1/responses` sanitized before routing | gated on `!target.is_remote()` |
| `openclaw` / `gemini` / `cline` / `kimi` / `copilot` | `PROVIDER_UNSUPPORTED` |
| `github-copilot` needs OAuth | `OAUTH_ONLY` |
| empty catalog cached as success | `from_json` errors when nothing is routable |
| CSRF via the daemon key | `is_cross_site` |
| key over a remote `LLMMAN_HOST` | `connects_over_loopback` refusal |
| Codex against a provider without Responses | `explain_missing_route` |

Two remain deliberately open, both stated in the code rather than papered over:

- **Windows `fs::rename`** — the report is factually wrong. `std::fs::rename` is documented as replacing an existing destination, via `MoveFileExW`/`SetFileInformationByHandle`.
- **Loopback reachability is not authentication** — agreed, and no longer implied anywhere. It bounds the blast radius; a shared machine's other local accounts can already reach a loopback daemon for everything else it does. Real authentication for `llmman serve` is a larger change than this PR.

CI is green across all ten legs, `cargo fmt --check` is clean, 377 lib tests pass, and the podman/no-default-features build works. Merging.

PR #319, opened 2026-08-27
fix: clear clippy warnings across the tree

onuralpszr:
Clears every clippy warning that had a one line fix: redundant closures, no-op casts and conversions, an unused parameter, a stray blank line, and an empty format literal. No behavior change, and main goes from 15 warnings down to 4.

### DIFF
--- a/src/cmd/list.rs
+++ b/src/cmd/list.rs
@@ -43,11 +43,10 @@ pub fn run(args: &ListArgs) -> anyhow::Result<()> {
         .max(4);
 
     println!(
-        "{:<name_w$}    {:<16}    {:<10}    {}",
+        "{:<name_w$}    {:<16}    {:<10}    MODIFIED",
         "NAME",
         "ID",
         "SIZE",
-        "MODIFIED",
         name_w = name_w,
     );
 
--- a/src/cmd/launch.rs
+++ b/src/cmd/launch.rs
@@ -45,7 +45,7 @@ pub fn run(args: &LaunchArgs) -> anyhow::Result<()> {
     let model = args
         .model
         .as_deref()
-        .map(|m| crate::shortnames::resolve_ollama_api(m))
+        .map(crate::shortnames::resolve_ollama_api)
         .unwrap_or_default();
 
     // Ensure serve is running (start it in background if needed), preloading
--- a/src/cmd/run.rs
+++ b/src/cmd/run.rs
@@ -736,8 +736,8 @@ mod unix_readline {
             raw.c_lflag &= !(libc::ECHO | libc::ECHONL | libc::ICANON | libc::ISIG | libc::IEXTEN);
             raw.c_cflag &= !(libc::CSIZE | libc::PARENB);
             raw.c_cflag |= libc::CS8;
-            raw.c_cc[libc::VMIN as usize] = 1;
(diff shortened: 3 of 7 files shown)


onuralpszr:
@ericcurtin done all single pr

ericcurtin:
Reviewed and verified locally on `b75e6b5`:

- `cargo fmt --all -- --check` — clean
- `cargo test --lib --bins` — 341 passed / 3 ignored, plus 13 `llmman-bench` tests
- `cargo clippy --all-targets` — down to **3** warnings (`if_same_then_else` in `src/cmd/run.rs:392`, and two `too_many_arguments`)

Every change is a genuine no-op:

- `list.rs`: the old trailing `{}` had no width spec, so inlining the literal `MODIFIED` produces byte-identical output.
- `run.rs`: `libc::VMIN`/`VTIME` are already `usize` in the `libc` crate, so the `as usize` casts were no-ops. Builds green on both Linux and Darwin legs.
- `serve.rs`: `reqwest::StatusCode::from(resp.status())` was an identity conversion.
- `modelpack.rs`: this one actually removes a real warning that main is emitting today — `warning: unused variable: store` shows up in the current main E2E build log.

Merging.

---

**Follow-ups** (not blockers, no need to hold this PR):

1. **`src/webui.rs` — the doc comment now attaches to the wrong item.** Dropping the blank line silences `empty_line_after_doc_comments`, but it makes a module-level description ("Each constant holds the raw gzip-compressed bytes for the corresponding file…") into the doc for `INDEX_HTML` alone. It should be an inner doc comment (`//!`) at the top of the file instead, which fixes the lint *and* keeps the docs correct.

2. **Nothing stops these from coming back.** CI has `build`, `test` and `e2e` jobs but no clippy job, which is why 15 warnings (and a plain rustc `unused_variables`) accumulated on main. Worth adding `cargo clippy --all-targets -- -D warnings` to the `test` job's `check-fmt` leg (the same Linux x86_64 leg that already owns `cargo fmt --check`, for the same "lints are platform-independent, run them once" reason). That needs the 3 remaining warnings dealt with first — see 3.

3. **The 3 survivors.** `if_same_then_else` at `src/cmd/run.rs:392` looks like the only one with a real question behind it: `multiline.is_some()` and `!paste_sb.is_empty()` both yield `". "`, so either they should be one branch or the paste prompt was meant to differ from the multiline one. Worth a look from whoever wrote the paste path before collapsing it. The two `too_many_arguments` (`serve.rs:1683` `spawn_llama_server`, `container.rs:232` `spawn`) share most of their parameter list and both want the same options struct, so they're one refactor, not two.

src/cmd/list.rs
```diff
@@ -43,11 +43,10 @@ pub fn run(args: &ListArgs) -> anyhow::Result<()> {
         .max(4);
 
     println!(
-        "{:<name_w$}    {:<16}    {:<10}    {}",
+        "{:<name_w$}    {:<16}    {:<10}    MODIFIED",
         "NAME",
         "ID",
         "SIZE",
-        "MODIFIED",
```
ericcurtin, line 50: In this case, I preferred our style prior to this, even though clippy disagrees (guess clippy sees this as a tiny micro-optimization, one that's not worth it in my eyes).

PR #320, opened 2026-08-27
fix(ci): run the unit tests that the bin target no longer holds

onuralpszr:
The unit test job ran `cargo test --bin llmman`, which has run 0 tests since the lib target was added: every `#[cfg(test)]` module moved into the lib, so the bin target has none left. `--lib --bins` runs all 341 of them again.

With the tests running, the Windows legs failed on the download marker: `touch` writes zero bytes, and a zero byte write does not move the mtime on Windows, so a long download looked stale and got stopped part way through. It now sets the mtime explicitly.

The daemon process group fix that was here before is dropped, #317 covers it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,8 +167,8 @@ jobs:
           if-no-files-found: error
 
   # --------------------------------------------------------------------------
-  # Unit tests: every #[cfg(test)] module under src/ (`cargo test --bin`,
-  # since this crate has no lib target — see Cargo.toml). These are pure-
+  # Unit tests: every #[cfg(test)] module under src/, which live in the
+  # lib target and the llmman-bench binary. These are pure-
   # Rust logic tests with no OS-specific behavior, but the crate itself
   # still links the Go shim via CGO, so a target-specific toolchain issue
   # (the exact kind the `build`/`e2e` jobs' own git history is full of —
@@ -288,8 +288,8 @@ jobs:
       # comment on cargo test compiling its own separate test binary,
       # which links against the same Go-shim-produced llmman_shim.lib the
       # main binary does and hit the exact same lld-link requirement.
-      - name: cargo test --bin llmman
-        run: cargo test --release --target ${{ matrix.target }} --bin llmman
+      - name: cargo test --lib --bins
+        run: cargo test --release --target ${{ matrix.target }} --lib --bins
         env:
           RUSTFLAGS: ${{ matrix.rustflags }}
 
--- a/src/llama_release.rs
+++ b/src/llama_release.rs
@@ -459,7 +459,12 @@ impl DownloadMarker {
     /// download from a crashed one whose Drop never ran.
     fn touch(&self) {
         if let Some(path) = &self.0 {
+            // Windows keeps the old mtime when a write is zero bytes,
+            // so set it explicitly.
             let _ = std::fs::write(path, b"");
+            if let Ok(file) = std::fs::File::options().write(true).open(path) {
+                let _ = file.set_modified(std::time::SystemTime::now());
+            }
         }
     }
 }


onuralpszr:
Working CI Example from fork: https://github.com/onuralpszr/llmman/actions/runs/33123225590

ericcurtin:
Checked out `4822e61` and ran it: `cargo fmt --check` clean, `cargo test --lib --bins` green (341 lib + 13 `llmman-bench`).

The diagnosis is right and I confirmed both halves independently.

**The CI target bug is real and worse than the description says.** `Cargo.toml` does have a `[lib]` target, so the workflow comment that `cargo test --bin llmman` exists "since this crate has no lib target" is simply stale. `src/main.rs` contains no `#[cfg(test)]` at all, so the `Unit tests` job has been running **zero tests on all five targets** while reporting green. `--lib --bins` picks up 341+13. That's the single most valuable thing in this PR.

**The E2E failure is real too.** Latest main run [33090803965](https://github.com/llmmanorg/llmman/actions/runs/33090803965): only the three Linux legs fail (x86_64 docker, x86_64 podman, aarch64 docker); macOS and Windows pass — exactly the procps-ng-vs-BSD split you'd predict. The log ends with:

```
[spawn_with_timeout] pid=12247 spawned: `llmman pull` against a daemon that dies at startup
##[error]The runner has received a shutdown signal.
test ensure_server_fails_fast_when_daemon_dies_at_startup ...
```

So: right bug, right root cause. But I don't think this branch should merge as-is, for four reasons.

---

### 1. It collides with #317, which fixes the same bug more completely

#317 (opened ~4h before this one) is the same kill(2) change. I trial-merged both: they **conflict in `src/daemon.rs`**, so one of them has to be rebased regardless.

More importantly, this PR only converts `signal_group`. `kill_daemon` (`src/daemon.rs:267`) still does:

```rust
let group = Command::new("kill").args([signal, &format!("-{pid}")]).status();
```

That is the identical procps-ng footgun, still live on this branch. And it's a *user-facing* path, not just a CI one: `kill_daemon` is reached from `stop_stale_daemon`, which runs whenever any command meets a daemon from an older install. On Ubuntu that means "upgrade llmman, run `llmman pull`, and SIGTERM every process your user owns" — then SIGKILL 5s later via the escalation. So this branch leaves the more damaging instance of the bug in place while fixing the one CI happens to trip over. #317 converts both.

### 2. No guard on the negative pid

```rust
unsafe { libc::kill(-(pid as libc::pid_t), signal) == 0 }
```

Two problems. `pid == 1` gives `kill(-1, sig)` — the exact user-wide broadcast this PR is fixing. `pid == 0` gives `kill(0, sig)` — the caller's own group. And `pid as libc::pid_t` is a lossy `u32 -> i32` cast: a value above `i32::MAX` wraps negative, and `-(negative)` then signals a single unrelated pid (or panics on `i32::MIN` in a debug build). Yes, `child.id()` won't realistically return those today, but this function is one refactor away from taking a pid parsed off the wire in `DaemonIdentity`, and the failure mode is "take down the machine". #317's `kill_group` has the `pid <= 1` and `i32::try_from` guards plus a unit test (`kill_group_refuses_pids_that_would_signal_everything`) that pins them with signal 0.

### 3. Nothing here is covered by a test that runs on a PR

`e2e` is gated on `github.event_name == 'push'`, so the only job that exercises `ensure_server_fails_fast_when_daemon_dies_at_startup` cannot run on this PR — it stays SKIPPED, as it is above. That means the daemon half of this change lands unverified and is first exercised on main.

The linked "working CI example from fork" doesn't close that gap: run [33123225590](https://github.com/onuralpszr/llmman/actions/runs/33123225590) is on `063acdea`, and I fetched that commit — it's upstream's `fix(ci): retry, verify, and hard-fail integration CLI installs (#304)`, with `src/daemon.rs` still calling `Command::new("kill")` and `ci.yml` still on `cargo test --bin llmman`. **None of this PR's changes are in that run.** It's evidence that E2E passed on an older commit, not that this branch fixes it.

The fix is unit tests that run in the `test` job. #317 already has the shape: spawn a `process_group(0)` target plus a control child in the harness's own group, assert the target dies and the control survives — a regression would fail an assertion instead of the runner.

### 4. Two changes are outside the stated scope

`DownloadMarker::touch` in `src/llama_release.rs` is not mentioned in the title, the description, or either commit message:

```rust
let _ = std::fs::write(path, b"");
if let Ok(file) = std::fs::File::options().write(true).open(path) {
    let _ = file.set_modified(std::time::SystemTime::now());
}
```

The premise ("writing zero bytes leaves the mtime alone on Windows") may well be right, but it's an untested claim about a platform, it's a second `open()` where `File::create(path)` + `set_modified` would do, and it silently changes when `download_in_progress()` reports a live download — which is what decides whether a 60s startup timeout kills the daemon or leaves it running. That deserves its own PR with a test against `marker_is_fresh`.

The `has_exited` / `owns_group` rework is at least related, but it's a behavioural change to startup failure handling riding along with a CI fix, and CodeRabbit's point on it stands: `waitid` returning `-1` is folded into `Ok(false)`, so a failed status check makes `ensure_server` sit out the full 60s timeout instead of reporting. `EINTR` needs a retry, anything else needs `last_os_error()`.

---

### What I'd like to see

the `DownloadMarker::touch` mtime fix**, on its own, with a test.

onuralpszr:
@ericcurtin I didn't see daemon PR truly, I revert it and leave that to you and I minimize mine to just cover test parts thank you.

ericcurtin:
This is exactly the right shape now — thanks for splitting it. Checked out `9dd36a6`: `cargo fmt --check` clean, `cargo test --lib --bins` green (341 lib + 13 `llmman-bench`), all five Unit test legs and both Windows legs green in CI.

I went and confirmed the Windows failure you describe rather than taking it on faith, because "CI change also touches `llama_release.rs`" is normally a scope-creep smell. It isn't here. Run [33112925304](https://github.com/llmmanorg/llmman/actions/runs/33112925304), the first push of this branch (before the marker fix), on `x86_64-pc-windows-msvc`:

```
test llama_release::tests::download_marker_creates_touches_and_removes_the_marker ... FAILED
thread '...' panicked at src\llama_release.rs:1098:9:
assertion failed: marker_is_fresh(&path, Duration::from_secs(60))
test result: FAILED. 333 passed; 1 failed; 3 ignored
```

Both Windows legs, same assertion — the one immediately after `marker.touch()`. So the two changes genuinely belong in one commit: enabling the tests is what exposes the bug, and the CI fix can't land green without it.

Worth spelling out how bad the dark spot was. The lib target arrived in `a4abf3c` (2026-08-25, 34 commits back). From that commit on, every `#[cfg(test)]` module moved into the lib, `cargo test --bin llmman` had nothing left to run, and the job reported green while executing **zero tests on all five targets**. The stale workflow comment asserting "this crate has no lib target — see Cargo.toml" is what kept it invisible; good catch fixing the comment and not just the command.

And the bug it was hiding is user-facing, not just a test artifact. `download_in_progress()` is what `ensure_server` consults at its 60s startup timeout to decide between "still fetching llama-server, leave the daemon alone" and "dead, stop it". With `touch()` never moving the mtime on Windows, a llama-server download longer than `DOWNLOAD_MARKER_STALE_AFTER` always looked stale, so the daemon got `stop_group`'d mid-download — and since the staging files are pid-specific with no resume, every retry restarted from zero. That's a Windows-only "large download never completes" loop that no one could have seen, because the test proving it never ran.

Merging.

---

**Follow-ups** (none blocking):

1. **Add a clippy job now that the tests are real.** Same argument as the dark unit tests: nothing enforces lints, which is how 15 clippy warnings plus a plain rustc `unused_variables` accumulated before #319. `cargo clippy --all-targets -- -D warnings` on the `check-fmt` leg (Linux x86_64, which already owns `cargo fmt --check` for the "lints are platform-independent, run once" reason). Three warnings need clearing first — see #319 for the list; `if_same_then_else` at `src/cmd/run.rs:392` is the only one with a real question behind it.

2. **`touch()` opens the file twice.** `fs::write` + a second `File::options().write(true).open()` can collapse into `File::create(path)` (which already creates-or-truncates, so it keeps the "recreate if a concurrent Drop removed it" property) followed by `set_modified`. Cosmetic, but it removes the reader's "why two opens?" pause.

3. **The Windows regression is covered incidentally, not deliberately.** `download_marker_creates_touches_and_removes_the_marker` only catches this because it backdates the mtime by 120s before calling `touch()`. Anyone who later "simplifies" that setup silently deletes the only coverage of the zero-byte-write semantics, on the one platform where it matters. Worth either a comment on the backdating step saying what it's really guarding, or a dedicated test named for the actual invariant (`touch_moves_the_mtime_even_when_the_marker_is_already_empty`).

4. **`--lib --bins` still excludes doctests and integration tests.** Zero doctests exist today (I checked — `cargo test --doc` runs 0; the two fenced blocks in `storage/oci.rs` and `cmd/resolve.rs` are `text`/`json`, not Rust), so there's no gap right now. But the first real doctest anyone writes will be skipped just as silently as these 341 were. Cheap insurance: append `--doc` as a second step, or switch to `--all-targets --doc`. `tests/launch_e2e.rs` is deliberately out of scope here since `e2e` owns it.

5. **34 commits landed with no unit-test coverage.** This PR proves at least one real bug slipped through in that window. CI is green now, so nothing else is currently broken, but it's worth knowing that's the blast radius if something odd turns up on Windows soon.

PR #323, opened 2026-08-29
fix(serve): pass --threads from cgroup CPU quota

### LINKED ISSUE #322: serve: llama-server threads ignore container CPU quota
## Problem

When llmman runs llama-server inside a container limited with [`--cpus`](https://docs.docker.com/engine/containers/resource_constraints/#cpu) (for example `docker run --cpus=2`), generation is much slower than the quota should allow. Measured below: 4.1x slower with a 2-CPU quota on a 16-core host, compared to running with a thread count matching the quota.

## Root cause

Docker's `--cpus` sets a CFS bandwidth quota (cgroup v2 `cpu.max`), not CPU affinity. It throttles time; it does not shrink the set of visible cores.

[`spawn_llama_server`](https://github.com/llmmanorg/llmman/blob/c95a5e19526e2ff411d376ec2f3b9c7370c6ec74/src/cmd/serve.rs#L1683) never passes `--threads`, so llama-server falls back to its own autodetection, which reads the affinity mask and still reports every host core. The result: more threads than the quota can serve, all fighting each other for the same time budget and stalling on CFS throttling.

## Reproduction

Measured on an AMD Ryzen 7 7840U (8 cores / 16 threads), kernel 7.1.6, Docker 29.2.1, cgroup v2, llmman 0.1.0 (5c5b936), llama-server b10621 CPU build.

```sh
# Any container with a CPU quota (the actual trigger: Docker --cpus is a
# cgroup time quota, not a cpuset/affinity restriction)
docker run -it --rm --cpus 2.0 --name llmman-repro debian:trixie-slim bash

# --- inside the container ---
apt-get update && apt-get install -y --no-install-recommends ca-certificates curl jq libgomp1
curl -fsSL https://raw.githubusercontent.com/llmmanorg/llmman/main/install.sh | sh
export PATH="$HOME/.local/bin:$PATH"

# baseline: no --threads override, llama-server autodetects host core count
llmman serve &
sleep 2
curl -s -X POST http://localhost:17434/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"hf.co/unsloth/Qwen3.5-2B-GGUF:Q4_K_M","messages":[{"role":"user","content":"Count from 1 to 20, one number per line, nothing else."}],"max_tokens":60,"stream":false}' \
  | jq .timings

llmman stop hf.co/unsloth/Qwen3.5-2B-GGUF:Q4_K_M
kill %1

# same request, only difference: LLAMA_ARG_THREADS set to match the quota
LLAMA_ARG_THREADS=2 llmman serve &
sleep 2
curl -s -X POST http://localhost:17434/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"hf.co/unsloth/Qwen3.5-2B-GGUF:Q4_K_M","messages":[{"role":"user","content":"Count from 1 to 20, one number per line, nothing else."}],"max_tokens":60,"stream":false}' \
  | jq .timings
```

Compare `predicted_per_second` (or raw `predicted_ms` / `predicted_n`) between the two `.timings` outputs. Five runs each after a warm-up request, same model, prompt, and `max_tokens`:

| Configuration | `predicted_per_second` (median of 5) | range |
|---|---|---|
| Default (autodetected threads) | 4.41 | 4.38 - 4.47 |
| `LLAMA_ARG_THREADS=2` | 17.89 | 17.83 - 17.92 |

The matched thread count is 4.1x faster on identical hardware, model, and prompt. The only difference between the two runs is the thread count.

## Proposed fix

Derive `--threads` from the cgroup quota when one is present, following the existing tunable pattern (pure parse function plus env wrapper, as with [`LLMMAN_NUM_PARALLEL`](https://github.com/llmmanorg/llmman/blob/c95a5e19526e2ff411d376ec2f3b9c7370c6ec74/src/cmd/serve.rs#L235)):

- Parse cgroup v2 `cpu.max` (`<quota> <period>` or `max <period>`) and compute `floor(quota / period)`, floored at 1 and capped at `available_parallelism`. Fall back to cgroup v1 `cpu.cfs_quota_us` / `cpu.cfs_period_us` (`-1` means unlimited).
- Skip entirely when `LLAMA_ARG_THREADS` is set: llama-server already reads it through plain env inheritance, and an explicit user choice must keep winning.
- Fail closed: `max`, unparseable content, or a missing `/sys/fs/cgroup` (rootless or nested containers) returns nothing and leaves llama-server's autodetection untouched. Linux only; a no-op elsewhere.

Known scope limit: fixed cgroup paths miss non-default layouts (`--cgroupns=host`, systemd scopes, a child cgroup whose ancestor holds the tighter limit). Every miss fails closed, so behavior there stays exactly as today. In practice the `available_parallelism` cap still corrects most of these, since Rust's std walks the cgroup hierarchy itself.

maxgio92:
Fixes #322.

Docker's `--cpus` sets a CFS bandwidth quota (cgroup v2 `cpu.max`), not CPU affinity. `spawn_llama_server` passed no `--threads`, so llama-server autodetected from host cores and its threads fought over the quota (measured 4.1x slowdown at `--cpus=2` on a 16-core host; repro and numbers in #322).

## What this does

- [`threads_from_env_or_host`](https://github.com/maxgio92/llmman/blob/79071b6/src/cmd/serve.rs#L258) passes `--threads` only when a CPU limit binds: when `std::thread::available_parallelism()` (which is `min(affinity, cgroup quota)`; std walks `/proc/self/cgroup` and the ancestor chain itself, v1 and v2) is below the online CPU count from `/sys/devices/system/cpu/online`. It then passes `available_parallelism()` itself.
- Unconstrained hosts get no flag: llama-server's own `cpu_get_num_math()` autodetection stays untouched.
- `LLAMA_ARG_THREADS` set in the environment wins: llmman passes nothing and llama-server reads the variable itself. The variable is documented in the serve env help and on `LLAMA_CPP_ENV_PASSTHROUGH_VARS`, so it also reaches a containerized llama-server under `--ociman`.
- The value resolves once at startup, lives on `Inner` beside the other tunables, is logged, and rides `LlamaOptions.threads`; `container::spawn` deliberately ignores it (the daemon's cgroup says nothing about the child container's).

## Fail-closed behavior

Every miss (unreadable or malformed `/sys` content, non-Linux) passes no flag and keeps autodetection. One accepted tradeoff, in the doc comment: a quota between physical cores and SMT threads (e.g. `--cpus=12` on an 8-core/16-thread host) passes 12 where autodetection would pick 8; correcting it would need a local topology walk that risks drifting from upstream's.

Follow-ups: #324 (CPU limit for the `--ociman` container), #328 (per-request `options.num_thread`, which should win over the startup-derived value).

## Testing

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and the full test suite pass. Table-driven unit tests cover the `/sys` CPU-list counting, including overflow rows.

Measured end to end with the reproduction from #322 (model `Qwen3.5-2B-GGUF:Q4_K_M`, same prompt and `max_tokens`; `predicted_per_second` from the response's `.timings`, median of 5 runs after a warm-up):

| Binary | Env | tokens/s |
|---|---|---|
| release 0.1.0 (5c5b936) | none | 4.41 |
| release 0.1.0 (5c5b936) | `LLAMA_ARG_THREADS=2` | 17.89 |
| this PR | none | 17.61 |

Verified on the final shape: `--cpus=2` derives `--threads 2`, an unconstrained host passes no flag, and a set `LLAMA_ARG_THREADS` is left untouched.


src/cmd/serve.rs
```diff
@@ -237,6 +237,78 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+    #[cfg(target_os = "linux")]
+    {
+        let available = std::thread::available_parallelism().ok()?.get() as u32;
+        // cgroup v2 first; fall back to the v1 cpu controller files.
+        if let Ok(cpu_max) = std::fs::read_to_string("/sys/fs/cgroup/cpu.max") {
+            return effective_threads_from_cgroup(&cpu_max, available);
+        }
+        let quota = std::fs::read_to_string("/sys/fs/cgroup/cpu/cpu.cfs_quota_us").ok()?;
+        let period = std::fs::read_to_string("/sys/fs/cgroup/cpu/cpu.cfs_period_us").ok()?;
+        effective_threads_from_cgroup_v1(&quota, &period, available)
```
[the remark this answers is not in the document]
maxgio92: Declining: the fixed paths are a deliberate scope decision, documented under "Fail-closed behavior" in the PR description; they resolve the quota in the target Docker --cpus case, and every other layout fails closed to None because the v2 root exposes no cpu.max and the v1 root quota reads -1. Over-subscription cannot happen either way: the result is capped with .min(available), and std available_parallelism already resolves /proc/self/cgroup and takes the minimum over ancestor cpu.max limits, so walking the hierarchy here would add parsing surface without changing any outcome.

maxgio92:
@coderabbitai the cpu.max extra-fields finding is valid: fixed in 998d2fa. Content with more than two fields now fails closed to None, and the malformed-input test covers both the finite and the max form.

src/cmd/serve.rs
```diff
@@ -237,6 +238,82 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
     (n != 0).then_some(n)
 }
 
+/// `--threads <n>` for every local `llama-server` spawn, derived from the
+/// container's CPU quota. Docker's `--cpus` sets a CFS bandwidth quota
+/// (cgroup `cpu.max`), not CPU affinity, so llama-server's own thread
+/// autodetection (`available_parallelism`, which reads the affinity mask)
```
ericcurtin: That is Rust's implementation, not llama.cpp's. Upstream uses `cpu_get_num_math()`, counting physical cores via `/sys/devices/system/cpu/cpu*/topology/thread_siblings` and falling back to `hardware_concurrency()`. The conclusion holds, since neither consults the CFS quota, but this comment is what the next reader will trust instead of the source.

src/cmd/serve.rs
```diff
@@ -237,6 +238,82 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+/// reports every host core and the resulting threads fight over the
+/// quota. `LLAMA_ARG_THREADS` set in the environment wins: llama-server
+/// reads it itself via plain env inheritance, so `None` here keeps that
+/// explicit choice untouched. Any missing/unreadable cgroup file also
+/// returns `None` (rootless or nested containers without /sys/fs/cgroup
+/// keep current behavior).
+fn threads_from_env_or_cgroup() -> Option<u32> {
```
ericcurtin: Reads env and the filesystem on every call for a value that cannot change. Move it to the constructor beside `num_parallel: num_parallel_from_env()`, store on `Inner`, and pass it down like every other spawn parameter.

src/cmd/serve.rs
```diff
@@ -237,6 +238,82 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+    )
+}
+
+/// `floor(quota / period)` with a floor of 1, capped at `available`;
+/// `None` for zero fields (nothing meaningful to compute).
+fn quota_threads(quota: u64, period: u64, available: u32) -> Option<u32> {
+    if quota == 0 || period == 0 {
+        return None;
+    }
+    let n = (quota / period).max(1).min(u64::from(available));
```
ericcurtin: `available` is the SMT count. It rarely binds with a quota present, but when it does (affinity narrower than the quota, or `--cpus` at or above the host core count) it hands over roughly double what this workload wants, since siblings contend for the same execution units. Upstream counts physical cores for that reason.

Capping at physical cores is another `/sys` topology read and stays a pure function. It also opens up a better unconstrained default later.

src/cmd/serve.rs
```diff
@@ -2559,6 +2642,7 @@ async fn ensure_model(
                     context_shift,
                     split_mode,
                     num_parallel,
+                    threads_from_env_or_cgroup(),
```
ericcurtin: Inside the OOM retry loop, so `/sys` is re-read per attempt for a value that cannot change. Read it off `state.0`, like `num_parallel` two arguments up.

ericcurtin (changes requested):
Convincing measurement, and the parse helpers are the right shape: pure functions, table-driven tests, fail-closed. My concerns are about where the decision is made, not the arithmetic.

1. Resolve once at startup. Every other tunable (`ctx_size`, `split_mode`, `num_parallel`, `max_queue`) is read from env in the constructor and stored on `Inner`. This one is called inline at serve.rs:2645, inside the OOM retry loop. Log the result too, so the next "my container is slow" report is a glance instead of a bisect.

2. Read `/proc/self/cgroup` rather than assuming the root. That closes `--cgroupns=host`, systemd scopes, and tighter ancestors, all currently listed as accepted misses. Roughly 15 lines, same test style. A miss silently restores the 4x slowdown.

3. `--ociman` (serve.rs:2618) gets no threads. Correct, since this daemon's cgroup says nothing about a child container's, but say so in the doc comment. `container::spawn` also sets no CPU limit on `docker run`, reproducing #322 one level down. Separate issue.

Not a blocker: thread count is per-model, not per-daemon. The plumbing already carrying `num_ctx` is its natural home.

maxgio92:
Thank you @ericcurtin. I'm going to address the review. Makes sense to me.

PS: Good catch on the cgroup sysfs, my bad 🤦🏻

src/cmd/serve.rs
maxgio92: Right! Good catch. Thanks

maxgio92:
@ericcurtin all three addressed, thanks for the thorough pass.

1. Startup: resolved once in addfb2d, stored on `Inner`, logged at startup, read from state at the spawn site.
2. Cgroup resolution: c56d32d parses `/proc/self/cgroup`, matches the mount from `/proc/mounts`, and walks the chain taking the strictest finite `cpu.max`. Same pure-function-plus-tests shape.
3. `--ociman`: rationale is now a doc comment on the spawn arm (0161e65). The missing CPU limit on the child container is filed as #324.

The per-model plumbing suggestion makes sense as a follow-up; left out here to keep this PR on the default.

src/cmd/serve.rs
maxgio92: Right, that was Rust's behavior, not llama.cpp's. Fixed in 95571b6: the comment now describes cpu_get_num_math() and keeps the conclusion that neither consults the quota.

src/cmd/serve.rs
maxgio92: Fixed in addfb2d. Resolved once at startup next to num_parallel_from_env(), stored on Inner, passed down like the other spawn parameters.

src/cmd/serve.rs
maxgio92: Fixed in 41e3fe6: the cap is now distinct physical cores from thread_siblings_list, falling back to available_parallelism when topology is unreadable. b70bf4e fixes a bug the first cut had: it probed cpu IDs from a quota-capped count, which undercounted cores on adjacent-sibling SMT hosts; the probe now uses the IDs in /sys/devices/system/cpu/present.

src/cmd/serve.rs
maxgio92: Fixed in addfb2d, it now reads state.0.threads like num_parallel.

src/cmd/serve.rs
```diff
@@ -58,6 +58,7 @@ Environment Variables:
       LLMMAN_TMPDIR                  Staging directory for llama-server release downloads
       LLAMA_ARG_FIT                  Enable llama.cpp automatic fit of unset memory options (default \"on\")
       LLAMA_ARG_FIT_TARGET           Target free VRAM margin per device for llama.cpp fit (MiB)
+      LLAMA_ARG_THREADS              Thread count for llama-server (default: CPU quota when containerized)
```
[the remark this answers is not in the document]
maxgio92: Valid: fixed in a5c2c93. LLAMA_ARG_THREADS is on LLAMA_CPP_ENV_PASSTHROUGH_VARS now, so the override reaches the containerized llama-server too.

src/cmd/serve.rs
```diff
@@ -237,6 +238,82 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+    #[cfg(target_os = "linux")]
+    {
+        let available = std::thread::available_parallelism().ok()?.get() as u32;
+        // cgroup v2 first; fall back to the v1 cpu controller files.
+        if let Ok(cpu_max) = std::fs::read_to_string("/sys/fs/cgroup/cpu.max") {
+            return effective_threads_from_cgroup(&cpu_max, available);
+        }
+        let quota = std::fs::read_to_string("/sys/fs/cgroup/cpu/cpu.cfs_quota_us").ok()?;
+        let period = std::fs::read_to_string("/sys/fs/cgroup/cpu/cpu.cfs_period_us").ok()?;
+        effective_threads_from_cgroup_v1(&quota, &period, available)
```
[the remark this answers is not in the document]
maxgio92: Valid: fixed in c56d32d. The v1 mount is matched by its controller list from /proc/self/cgroup and /proc/mounts, so cpu,cpuacct co-mounts resolve in either order, with a test covering both.

src/cmd/serve.rs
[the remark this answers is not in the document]
maxgio92: Revisiting this: the resolution landed in c56d32d. The process cgroup now comes from /proc/self/cgroup joined with the mount, with an ancestor walk taking the strictest limit.

src/cmd/serve.rs
```diff
@@ -237,6 +238,289 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+    #[cfg(target_os = "linux")]
+    {
+        let available = std::thread::available_parallelism().ok()?.get() as u32;
+        // Cap at physical cores, the same scope llama-server's own
+        // cpu_get_num_math() default targets: on an SMT host a quota of
+        // e.g. 16 CPUs still shouldn't oversubscribe 8 physical cores.
+        // `available` stays a bound of its own: it carries the affinity
+        // mask and std's quota detection, which topology knows nothing
+        // about.
+        let cap = physical_core_count_fs().unwrap_or(available).min(available);
```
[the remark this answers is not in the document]
maxgio92: Partly agreed: the hybrid x86 half is fixed in 1b94ff5. The cap now prefers the P-core list from /sys/devices/cpu_core/cpus when present, falling back to the all-physical-core count elsewhere. POWER SMT-2 parity stays out: it needs per-arch topology parsing for a case where the quota rarely binds, and available_parallelism still bounds the result.

src/cmd/serve.rs
```diff
@@ -237,6 +238,289 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+    let present = std::fs::read_to_string("/sys/devices/system/cpu/present").ok()?;
+    let sibling_lists: Vec<String> = cpu_id_list(&present)?
+        .into_iter()
+        .filter_map(|i| {
+            std::fs::read_to_string(format!(
+                "/sys/devices/system/cpu/cpu{i}/topology/thread_siblings_list"
+            ))
+            .ok()
+        })
+        .collect();
```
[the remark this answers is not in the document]
maxgio92: The skip is deliberate; the doc comment was wrong and is fixed in d813dee. Offline CPUs stay listed in present while the kernel unregisters their topology directory, and an offline core contributes no execution capacity, so excluding it is more accurate than falling back to available_parallelism (the SMT count, which over-provisions). Only the all-unreadable case falls back.

src/cmd/serve.rs
```diff
@@ -237,6 +238,289 @@ fn parse_num_parallel(value: Option<&str>) -> Option<u32> {
...
+}
+
+/// The strictest (minimum) thread count across a chain of `cpu.max`
+/// contents. The kernel enforces every level, so the tightest quota
+/// wins regardless of where in the chain it sits; `None` when no level
+/// has a finite quota.
+fn strictest_threads_v2<'a>(cpu_maxes: impl IntoIterator<Item = &'a str>, cap: u32) -> Option<u32> {
+    cpu_maxes
+        .into_iter()
+        .filter_map(|cpu_max| effective_threads_from_cgroup(cpu_max, cap))
```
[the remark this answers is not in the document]
maxgio92: Valid: fixed in 1800d32. cpu.max and the v1 cfs files now parse to a tri-state (finite, unlimited, malformed), and any malformed entry fails the whole chain closed, including "max <bad-period>" and v1 quotas other than -1. Missing files stay skippable: the v2 root has no cpu.max at all.

ericcurtin (changes requested):
`std::thread::available_parallelism()` already returns `min(affinity, cgroup quota)`, walking `/proc/self/cgroup` and the ancestor chain for v1 and v2 (`std/src/sys/thread/unix.rs`, `cgroups::quota()`).

So `cap` never exceeds the chain quota, and `threads_from_cgroup_fs(cap)` always returns exactly `cap`. The 300 lines of cgroup parsing only decide `Some` vs `None`. Keep the physical-core cap, drop the rest:

```rust
fn threads_from_env_or_cgroup() -> Option<u32> {
    if std::env::var_os("LLAMA_ARG_THREADS").is_some() {
        return None;
    }
    let allowed = std::thread::available_parallelism().ok()?.get() as u32;
    Some(math_core_count_fs().unwrap_or(allowed).min(allowed))
}
```

Also: CPU discovery belongs in its own module beside `hostgpu.rs`. Follow-up: honor `options.num_thread` per request.

maxgio92:
Agreed: done in f760fbb. The cgroup parsing is gone and the function is your exact shape, renamed to `threads_from_env_or_host` since it no longer touches cgroups. Verified in a container: `--cpus=2` derives `--threads 2` through std alone, and an unconstrained 8-core host derives 8, matching llama-server's own default.

One consequence stated in the commit: `--threads` is now always passed on Linux unless `LLAMA_ARG_THREADS` is set, not only under a quota. And one caveat kept in the doc comment: std's walk assumes the standard `/sys/fs/cgroup` mount and is best-effort on malformed levels; a miss falls back to min(math cores, affinity), the pre-feature behavior.

CPU discovery moved to `src/hostcpu.rs` in 3ea44f9. Filed the per-request `options.num_thread` follow-up as #328.

(Edit: commit ids repointed after a rebase onto current main; content unchanged.)

ericcurtin (commented):
CI is green and the earlier points are all addressed. The conflict with `main` is a blocker.

#327 folded the spawn arguments into `container::LlamaOptions`, so `spawn_llama_server` now takes `opts` instead of the positional list. To resolve: add `threads: Option<u32>` to `LlamaOptions`, set it from `state.0.threads` where `llama_opts` is built, and destructure it as `threads: _` in `container::spawn` with your "the daemon's cgroup says nothing about the child container's" rationale moved onto that field. Both conflict hunks collapse into that.
Follow-ons after merge:
- #324 CPU limit for the `--ociman` container.
- #328 per-request `options.num_thread`. That's the knob most users will reach for first, so it shouldn't stay env-var-only.

maxgio92:
Conflict resolved, in your shape: 8f81a23 carries `threads` as a `LlamaOptions` field, set where `llama_opts` is built, and `container::spawn` destructures it as `threads: _` with the rationale moved onto the field. The branch is rebased on current main and CI is green across all targets.

Agreed on the follow-ons: #324 and #328 after merge, and #328 first since per-request `options.num_thread` is the knob users will actually reach for.

ericcurtin:
llama-server already counts physical cores and honors the affinity mask. The only thing it misses is the cgroup CPU quota, so that is the only case worth correcting.

Suggestion: pass `--threads` only when the quota actually binds, i.e. `available_parallelism()` is below the online CPU count, and pass `available_parallelism()` then. Otherwise pass nothing and leave autodetection alone. That deletes `hostcpu.rs` entirely and removes the risk of the local topology walk drifting from the upstream `cpu_get_num_math()` it approximates (P-core list from /sys vs cpuid plus thread_siblings).

`num_thread` in request options is the standard knob for this. It should land (#328) and win over anything derived at startup.

maxgio92:
Done in 79071b6: `--threads` is passed only when `available_parallelism()` is below the online CPU count, and it carries `available_parallelism()` itself. Unconstrained hosts get no flag, so autodetection stays untouched. `hostcpu.rs` is deleted; the PR is now +133/-1 against main. CI is green and verified in containers: `--cpus=2` passes `--threads 2`, unconstrained passes nothing.

One accepted tradeoff, noted in the doc comment: a quota between physical cores and SMT threads (e.g. `--cpus=12` on an 8-core/16-thread host) passes 12 where autodetection would pick 8. Correcting that needs the topology walk we just deleted, so it stays with the quota.

Agreed on #328 winning over the startup-derived value once it lands.

ericcurtin:
Merged.

Follow-ons:
- #328 first: per-request `options.num_thread` should override the startup value.
- #324 for the container CPU limit.
- Track one more: a cpuset-only limit yields the mask width, not the physical cores inside the mask, so an SMT host can oversubscribe 2x.

maxgio92:
Thanks. I'm going to track the cpuset-only case in a separated issue

PR #329, opened 2026-08-30
fix: default-tag unload references and gate unload on a request keep_alive

ricky-chaoju:
Follow-up to #326

The unload paths never ran `default_tag`, while the load path gets it from `ensure_model`. `canonical_ref` supplies the tag only while the model is still in the store, so once it is not, a tagless unload looked for `docker.io/ai/m` while the model ran under `docker.io/ai/m:latest`, and the remove missed while the reply still said "unload". The gate also tested the resolved `keep_alive`, which falls back to the daemon default, so under `LLMMAN_KEEP_ALIVE=0` a plain preload evicted the model it was asking to warm. Both handlers now go through `unload_key` and `is_explicit_unload` so they cannot drift apart again.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -2131,6 +2131,25 @@ fn canonical_ref(store_path: &std::path::Path, model_ref: &str) -> String {
         .unwrap_or_else(|| model_ref.to_owned())
 }
 
+/// The `mgr.running` key an unload request names, resolved through the
+/// same three steps `ensure_model` uses to build the key it inserts
+/// under: `resolve_ollama_api`, then `default_tag`, then `canonical_ref`.
+///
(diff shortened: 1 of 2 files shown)


ericcurtin:
LGTM. Matches the upstream contract exactly: unload only when the request itself carries `keep_alive: 0`, and the runner key resolved the same way the load path resolves it.

Follow-ons:

- `unload_key` skips the remote-target check `ensure_model` does first. No-op today since a remote ref never lands in `running`, but it is the one step the two paths still do not share.
- `stop` still does a `ps` precheck and then unloads. Sending the unload and turning a "not found" error into the message drops a round trip and the race between the two calls.
- `stop` now prints the `ps` name rather than what the user typed. Printing nothing on success is the usual behavior.

PR #332, opened 2026-08-31
feat(launch): add Pi integration

### LINKED ISSUE #330: Add "llmman launch pi"

w3lld1:
## Summary

- I added Pi to the `llmman launch` integration registry and dispatch path.
- I configure Pi through its documented `models.json` custom-provider format, preserve unrelated providers, and pass provider credentials through the child environment instead of writing them to disk.
- I added focused config regression tests plus a real Pi launch case to the existing cross-platform E2E matrix.

## Validation

- `cargo fmt --all -- --check`
- `cargo clippy --release --all-targets -- -D warnings`
- `cargo test --release --lib --bins` (443 passed, 3 ignored)
- I exercised the built `llmman launch pi` path against a local test daemon and verified the selected model, forwarded arguments, credential environment, and preserved Pi config.
- I verified the generated provider config with Pi 0.84.4 model discovery. I left the real model/`llama-server` Pi E2E run to the existing CI matrix because `llama-server` is not installed locally.

Fixes #330

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -422,7 +422,7 @@ jobs:
         run: go test -tags podman,containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper ./...
 
   # --------------------------------------------------------------------------
-  # E2E: `llmman launch <claude|opencode|codex|hermes|openclaw> --model
+  # E2E: `llmman launch <claude|opencode|pi|codex|hermes|openclaw> --model
   # qwen3.5:0.8b` (tests/launch_e2e.rs) against a real `llmman serve`, a
   # real pulled model, a real llama-server, and the real third-party CLIs
   # — not mocks. Mirrors the `build` job's own 5-target matrix exactly
@@ -631,7 +631,7 @@ jobs:
       # multi-line/array syntax; ubuntu/macOS already default to bash.
       #
       # Unlike hermes/mlx_lm.server (legitimately missing on some
-      # platforms), these four are plain npm packages expected to work
+      # platforms), these five are plain npm packages expected to work
       # everywhere — so install_cli fails the job, rather than letting
       # tests/launch_e2e.rs's on_path check silently skip, when one never
       # becomes a working binary.
@@ -644,7 +644,7 @@ jobs:
       # designed (npm tolerates one failed platform-tarball fetch, here
       # a registry blip, instead of failing the install), just not
       # detectable by a mere on_path/PATH check. Retrying rides that out.
-      - name: Install integration CLIs (claude, opencode, codex, openclaw)
+      - name: Install integration CLIs (claude, opencode, pi, codex, openclaw)
         shell: bash
         run: |
           failed_clis=()
@@ -678,6 +678,7 @@ jobs:
 
           install_cli "@anthropic-ai/claude-code" "@anthropic-ai/claude-code" "claude"
           install_cli "opencode-ai" "opencode-ai" "opencode"
+          install_cli "@earendil-works/pi-coding-agent" "@earendil-works/pi-coding-agent" "pi"
           install_cli "@openai/codex" "@openai/codex" "codex"
           # Pinned (not @latest): tests/launch_e2e.rs's own openclaw
           # invocation was verified against this exact release — its
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ Point an integration at a model in one step. `llmman launch` starts `serve` in t
 llmman launch claude --model gemma4
 ```
 
-Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
+Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode, Pi, and others) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
 
 Short names work wherever a model reference is accepted.
 
--- a/src/cmd/launch.rs
+++ b/src/cmd/launch.rs
@@ -382,6 +382,12 @@ const INTEGRATIONS: &[Integration] = &[
         install_hint: "https://opencode.ai",
     },
     Integration {
+        name: "pi",
+        description: "Pi coding agent",
+        binary: "pi",
+        install_hint: "npm install -g --ignore-scripts @earendil-works/pi-coding-agent",
+    },
+    Integration {
         name: "codex",
         description: "OpenAI Codex CLI",
         binary: "codex",
@@ -500,6 +506,7 @@ fn launch(name: &str, model: &str, api_key: &str, extra_args: &[String]) -> anyh
     match name.to_lowercase().as_str() {
         "claude" => launch_claude(model, api_key, extra_args),
         "opencode" => launch_opencode(model, api_key, extra_args),
+        "pi" => launch_pi(model, api_key, extra_args),
         "codex" => launch_codex(model, api_key, extra_args),
         "cline" => launch_simple("cline", "cline is not installed: npm install -g cline", model, extra_args),
         "aider" => launch_aider(model, api_key, extra_args),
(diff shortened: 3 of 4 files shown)


tests/launch_e2e.rs
```diff
@@ -787,6 +787,28 @@ fn launch_opencode_with_model() {
     );
 }
 
+#[test]
+fn launch_pi_with_model() {
+    eprintln!("[test] launch_pi_with_model: acquiring SERIAL");
+    let _guard = lock_serial();
+    eprintln!("[test] launch_pi_with_model: acquired SERIAL");
```
ericcurtin, line 794: Can we remove both these skips, so that is llama.cpp and pi accidentally stops getting installed in CI, we know.

README.md
```diff
@@ -141,7 +141,7 @@ Point an integration at a model in one step. `llmman launch` starts `serve` in t
 llmman launch claude --model gemma4
 ```
 
-Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
+Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode, Pi, and others) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
```
ericcurtin, line 144: Can we remove the brackets, so we don't have to keep this upto date everytime we add an integration. Or leave it as it was.

src/cmd/launch.rs
```diff
@@ -583,6 +590,84 @@ fn opencode_config(model: &str, api_key: &str) -> String {
...
+        .entry("providers")
+        .or_insert_with(|| serde_json::json!({}))
+        .as_object_mut()
+        .context("pi models.json providers must be an object")?;
+    providers.insert(
+        "llmman".to_string(),
+        serde_json::json!({
+            "baseUrl": format!("{server}/v1"),
+            "api": "openai-completions",
+            "apiKey": "$LLMMAN_PI_API_KEY",
```
ericcurtin, line 660: Write the placeholder key literally, like the codex/hermes configs do. Use the env ref only when `--provider` passes a real key, otherwise a plain `pi` run outside launch sees an unset var and the provider is unusable.

src/cmd/launch.rs
```diff
@@ -583,6 +590,84 @@ fn opencode_config(model: &str, api_key: &str) -> String {
...
+            "compat": {
+                "supportsDeveloperRole": false,
+                "supportsReasoningEffort": false
+            },
+            "models": [{ "id": model }]
```
ericcurtin, line 665: Set `contextWindow`. pi defaults to 128000, serve usually loads far less, so requests overflow. serve already knows it (`query_context_length`). Same for `input`/`reasoning` from model capabilities.

src/cmd/launch.rs
```diff
@@ -583,6 +590,84 @@ fn opencode_config(model: &str, api_key: &str) -> String {
...
+    };
+    let root = root
+        .as_object_mut()
+        .context("pi models.json root must be an object")?;
+    let providers = root
+        .entry("providers")
+        .or_insert_with(|| serde_json::json!({}))
+        .as_object_mut()
+        .context("pi models.json providers must be an object")?;
+    providers.insert(
```
ericcurtin, line 655: Replacing the whole entry drops models a user added under this provider. Tag the entries llmman writes and preserve the rest.

src/cmd/launch.rs
```diff
@@ -583,6 +590,84 @@ fn opencode_config(model: &str, api_key: &str) -> String {
...
+    if let Ok(dir) = std::env::var("PI_CODING_AGENT_DIR") {
+        if !dir.trim().is_empty() {
+            return Ok(PathBuf::from(dir));
+        }
+    }
+    let home = dirs::home_dir().context("no home directory")?;
+    Ok(home.join(".pi").join("agent"))
+}
+
+fn write_pi_config(model: &str) -> anyhow::Result<()> {
```
ericcurtin, line 625: Skip the write when the rendered contents are unchanged, like `write_codex_config`.

src/cmd/launch.rs
```diff
@@ -381,6 +381,12 @@ const INTEGRATIONS: &[Integration] = &[
         binary: "opencode",
         install_hint: "https://opencode.ai",
     },
+    Integration {
+        name: "pi",
+        description: "Pi coding agent",
+        binary: "pi",
+        install_hint: "npm install -g --ignore-scripts @earendil-works/pi-coding-agent",
```
ericcurtin, line 388: Drop `--ignore-scripts`, the package has no install scripts. Keeps the hint shaped like the others.

ericcurtin (commented):
Nice shape overall, tests included. A few things to fix before merge, see inline.

.github/workflows/ci.yml
```diff
@@ -422,7 +422,7 @@ jobs:
         run: go test -tags podman,containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper ./...
 
   # --------------------------------------------------------------------------
-  # E2E: `llmman launch <claude|opencode|codex|hermes|openclaw> --model
+  # E2E: `llmman launch <claude|opencode|pi|codex|hermes|openclaw> --model
```
ericcurtin, line 425: Job `name:` on line 453 still omits pi.

ericcurtin:
There are conflicts to fix now

ericcurtin (commented):
Conflicts: rebase on main. `Integration` lost its `install_hint` field in #337, so drop that field and the install hint in the `launch_pi` error too.

- Selection rests on `--model llmman/<model>`. Write `~/.pi/agent/settings.json` with `defaultProvider` and `defaultModel` instead, which is how that config is meant to pick a model.
- `pi_config` replaces the whole `llmman` provider entry, so any model a user added under it is lost. Keep entries that lack a marker key (e.g. `_launch: true` on the ones you write) and rebuild only your own, the same way this already preserves unrelated providers.
- Drop the `on_path("pi")` skip in the e2e test. `install_cli` in ci.yml already fails the job when pi never becomes a working binary, and the skip hides exactly that.
- A sixth serialized launch test needs the e2e job's `timeout-minutes` raised as well. See what #355 did.
- Storing `$LLMMAN_PI_API_KEY` instead of the key itself is good, keep that.

ericcurtin (commented):
Conflicts with main, and the base has moved a long way. `install_hint` was removed from `Integration` in #337, so the field and the npm hint in `launch_pi`'s error both have to go. `--provider` landed in #333 with `PROVIDER_UNSUPPORTED` and `PROVIDER_NEEDS_DAEMON_KEY`; confirm `llmman launch pi --provider ...` works, since the key travels in `LLMMAN_PI_API_KEY` rather than a file. Rebase and re-run the unit tests.

On the implementation:

1. `pi_config` replaces `providers.llmman` wholesale, so its `models` array only ever holds the model of the current launch. Launch two models and the first is gone. Merge the array instead: mark llmman's own entries with a flag, keep everything else, and rebuild only the marked ones.
2. Nothing writes `settings.json`. Set `defaultProvider` and `defaultModel` there so a plain `pi` afterwards still reaches the daemon, rather than relying only on `--model llmman/<model>` on argv.
3. The model entry carries only `id`. Add `input` (text plus image for a vision model), `reasoning` for a thinking model, and `contextWindow`. Without them a vision model silently loses image input.
4. `write_pi_config` overwrites the user's `models.json` with no backup. Write a `.bak` first.
5. This is the seventh e2e launch test but `timeout-minutes: 310` and the "6 tests run serialized" comment above it are unchanged. Bump both.
6. `run_launch` gives each attempt a fresh HOME but does not clear `PI_CODING_AGENT_DIR`, so a developer with it exported gets their real config written. Add an `env_remove` for it.

ericcurtin:
Converting this to draft until it's ready for review again

ericcurtin:
@w3lld1 are you planning on continuing with this one?

ricky-chaoju:
@ericcurtin the author never came back, so I picked this up. Rebased off main and redid it against the current launch code. Your eight points are in.

Two things I did differently: kept the `on_path("pi")` skip, since every other launch test still has one; and `--provider` passes the key on argv with pi's `--api-key` rather than an env ref, so it stays off disk. The stored `apiKey` is the literal placeholder like you said.

ericcurtin:
@ricky-chaoju yeah that's fine... Eventually we are gonna start closing PRs that have been inactive for ~1 month (time to be decided)... Can be re-opened if necessary, reviewing now.

src/cmd/launch.rs
```diff
@@ -852,6 +865,232 @@ fn opencode_config(
...
+    vision: bool,
+    context_length: Option<u64>,
+    extra_args: &[String],
+) -> anyhow::Result<()> {
+    let bin = find_on_path("pi").ok_or_else(|| anyhow::anyhow!("pi is not installed"))?;
+
+    let effective_model = if model.is_empty() { "default" } else { model };
+    write_pi_config(effective_model, thinking, vision, context_length)?;
+
+    let mut args = vec![
```
ericcurtin: Drop the injected `--model`/`--api-key`; run `pi` with only the user's args. `settings.json` already selects the model, and a `--provider` key on argv shows up in `ps`. Add `pi` to `PROVIDER_NEEDS_DAEMON_KEY` (like hermes/cline) so the key stays with the daemon and the file keeps the placeholder.

src/cmd/launch.rs
```diff
@@ -852,6 +865,232 @@ fn opencode_config(
...
+    model: &str,
+    api_key: &str,
+    thinking: Option<&ThinkingControls>,
+    vision: bool,
+    context_length: Option<u64>,
+    extra_args: &[String],
+) -> anyhow::Result<()> {
+    let bin = find_on_path("pi").ok_or_else(|| anyhow::anyhow!("pi is not installed"))?;
+
+    let effective_model = if model.is_empty() { "default" } else { model };
```
ericcurtin: Add `pi` to `MODEL_REQUIRED` instead of writing a fake `default` model into models.json/settings.json.

src/cmd/launch.rs
```diff
@@ -852,6 +865,232 @@ fn opencode_config(
...
+            models
+                .iter()
+                .filter(|m| m.get("id") != entry.get("id"))
+                .cloned()
+                .collect()
+        })
+        .unwrap_or_default();
+    let mut models = kept;
+    models.push(entry.clone());
+    root["providers"][PI_PROVIDER] = serde_json::json!({
```
ericcurtin: This replaces the whole `llmman` provider object, so user keys like `headers` under it are lost. Merge into the existing object: always set `baseUrl`, set `api`/`apiKey` only when absent, keep the rest.

src/cmd/launch.rs
```diff
@@ -852,6 +865,232 @@ fn opencode_config(
...
+    let input: &[&str] = if vision {
+        &["text", "image"]
+    } else {
+        &["text"]
+    };
+    let mut entry = serde_json::json!({
+        "id": model,
+        "input": input,
+        "reasoning": thinking.is_some_and(|t| t.thinks),
+        PI_MARKER: true,
```
ericcurtin: Nit: `_llmman` is written but never read. Either use it (rebuild only marked entries) or drop it.

ericcurtin (changes requested):
Thanks for picking this up. Conflicts with main now (#520 landed: ci.yml, README.md, launch.rs, launch_e2e.rs) — please rebase; the launch-test count in ci.yml becomes 12. A few things on the shape, inline. Setting to draft; mark ready when done and I'll re-review.

src/cmd/launch.rs
```diff
@@ -921,6 +927,235 @@ fn opencode_config(
...
+    write_pi_json(&dir.join("settings.json"), |existing| {
+        pi_settings_merged(existing, model)
+    })
+}
+
+/// Reads `path`, hands the parsed object to `merge`, and writes the result
+/// back. See [`write_qwen_settings_at`], which this follows step for step;
+/// the one addition is the BOM, which pi strips (`stripBom` before
+/// `JSON.parse`) and `serde_json` rejects.
+fn write_pi_json(
```
ericcurtin: This is `write_qwen_settings_at` minus the BOM strip. Extract one `write_json_merged(path, label, merge)` and call it from both; the BOM strip is harmless for qwen too.

src/cmd/launch.rs
```diff
@@ -921,6 +927,235 @@ fn opencode_config(
...
+}
+
+/// What node's `os.homedir()` would answer, which is where pi looks.
+///
+/// Per node's own `os` documentation it is not one order across both:
+/// POSIX reads `$HOME` and falls back to the effective UID, Windows reads
+/// `%USERPROFILE%` and falls back to the profile directory. Reading the
+/// other platform's variable would send llmman's write somewhere pi never
+/// reads whenever the two disagree.
+fn node_home() -> anyhow::Result<PathBuf> {
```
ericcurtin: Drop `node_home`/`expand_home`; do what `qwen_home` does: `dirs::home_dir()` + the existing `expand_tilde`. On Windows `dirs` already reads the profile dir, and the e2e test pins `PI_CODING_AGENT_DIR` anyway.

src/cmd/launch.rs
```diff
@@ -921,6 +927,235 @@ fn opencode_config(
...
+    provider.insert("baseUrl".into(), serde_json::json!(format!("{server}/v1")));
+    provider
+        .entry("api")
+        .or_insert_with(|| serde_json::json!("openai-completions"));
+    // A literal, not a `"$VAR"` reference pi would interpolate: see
+    // launch_pi's own doc comment.
+    provider
+        .entry("apiKey")
+        .or_insert_with(|| serde_json::json!("llmman"));
+    provider.entry("compat").or_insert_with(|| {
```
ericcurtin: Drop `compat`. The daemon honours `reasoning_effort`, so `supportsReasoningEffort: false` switches off a working feature. Just `baseUrl`, `api`, `apiKey`, `models`.

src/cmd/launch.rs
```diff
@@ -921,6 +927,235 @@ fn opencode_config(
...
+) -> serde_json::Value {
+    let input: &[&str] = if vision {
+        &["text", "image"]
+    } else {
+        &["text"]
+    };
+    let mut entry = serde_json::json!({
+        "id": model,
+        "input": input,
+        "reasoning": thinking.is_some_and(|t| t.thinks),
```
ericcurtin: Nit: only set `reasoning` when true; omit otherwise.

ericcurtin (changes requested):
Thanks, the shape is right now and the merge/settings/e2e points are all in. Three de-dup/parity items inline, then this is good to go. Setting to draft; mark ready when done and I'll re-review.

ericcurtin (approved):
All three items from the last round are in, and the shared home-dir helper with `cline_dir` is a better de-dup than what I asked for. models.json/settings.json shapes look right. Merging.

ericcurtin:
Merged, thanks @ricky-chaoju (and @w3lld1 for the original). Follow-ons if anyone wants them, each its own PR:

1. Offer to `npm install -g @earendil-works/pi-coding-agent` when `pi` is missing, after a confirm prompt — same as `launch cline` already does.
2. Detect the legacy `@mariozechner/pi-coding-agent` package (walk from the `pi` binary to its package.json) and migrate to the current one instead of launching the stale install.
3. When a model entry in models.json lacks `contextWindow`, rebuild it on the next launch rather than leaving it stale.
4. A `--provider` e2e case for pi, since its key path (daemon-held) differs from the others.

PR #338, opened 2026-08-31
refactor(serve): use AppError for pull/push 400s

maxgio92:
handle_pull and handle_push hand-rolled 400 JSON bodies instead of using the AppError type introduced for the other reference-rejection sites in serve.rs.

Routes both through AppError::status and AppError::bad_request, matching the pattern the other five call sites already use.

No status code or JSON body change; same {"error": "..."} shape as before.

Following #303.


ericcurtin:
Merged.

Follow-on: the push handler's "model not found" a few lines below still hand-builds its body. `AppError::status(StatusCode::NOT_FOUND, ...)` gives it the same shape as the two 400s.

PR #348, opened 2026-08-31
feat(launch): add Factory Droid integration

### LINKED ISSUE #345: Add "llmman launch droid"

mah1104ahm:
Add Factory Droid to `llmman launch`, backed by the local daemon. The launcher preserves unrelated Factory settings and selects the managed model through `sessionDefaultSettings`, without a Droid `--model` override. Managed entries are identified by the placeholder credential plus the expected llmman display name, ID prefix and numeric index, so an unrelated model using the same dummy key is preserved.

Settings updates use a backup, atomic replacement and a lock between cooperating llmman launches. Image flags follow model capabilities; both flags are removed when capabilities are unknown. The workflow installs pinned Droid 0.208.1 and exercises the real CLI through the suite's existing retry/tolerant response helper.

Validation on macOS arm64 for `e56b1a0`:

- `cargo fmt --all -- --check` and `git diff --check` passed.
- `cargo test --release --lib cmd::launch::tests --offline`: 22 passed, including preservation of a user model with the placeholder key and removal of stale image flags.
- `cargo test --release --test launch_e2e launch_droid_with_model --offline -- --nocapture --test-threads=1`: 1 passed with real Droid 0.208.1, CI-pinned llama.cpp b10293 and qwen3.5:0.8b. Warm-up succeeded; the Droid invocation completed in 15.29 seconds. The shared helper tolerates stochastic response wording, so this is not a strict content or cross-platform guarantee.

This PR is ready for upstream platform checks and maintainer review. The full library/binary suite and Clippy were not rerun locally for this follow-up.

Closes #345

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -716,6 +716,7 @@ jobs:
           install_cli "@anthropic-ai/claude-code" "@anthropic-ai/claude-code" "claude"
           install_cli "opencode-ai" "opencode-ai" "opencode"
           install_cli "@openai/codex" "@openai/codex" "codex"
+          install_cli "droid@0.208.1" "droid" "droid"
           # Pinned (not @latest): tests/launch_e2e.rs's qwen test relies on
           # this release's tool set under --safe-mode and on the
           # report_findings schema it excludes by name; a newer release can
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2032,6 +2032,7 @@ dependencies = [
  "serde_json",
  "sha2 0.10.9",
  "tar",
+ "tempfile",
  "tokio",
  "tokio-util",
  "toml",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ flate2     = "1"
 zip        = { version = "2", default-features = false, features = ["deflate"] }
 libloading = "0.8"
 multer     = "3"
+tempfile   = "3"
 
 # Real Xet-protocol downloads for HuggingFace files too large for the
 # plain-HTTP path to reliably serve — see src/xet_fetch.rs. Its
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Point an integration at a model in one step. `llmman launch` starts `serve` in t
 llmman launch claude --model gemma4
 ```
 
-Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
+Run `llmman launch` with no arguments to list the supported integrations (Claude Code, OpenCode, Factory Droid, and others) and whether each is installed. Any extra arguments after `--` are forwarded to the integration's own CLI.
 
 Short names work wherever a model reference is accepted.
 
@@ -305,4 +305,3 @@ Uses [`github.com/podman-container-tools/container-libs`](https://github.com/pod
 ```
 cargo build --release --no-default-features --features podman
 ```
-
--- a/README.md
+++ b/README.md
@@ -253,8 +253,7 @@ The API key is read from the variable models.dev names for that provider
 and travels per request, never to disk. `droid` and `hermes` are the
 exceptions: llmman configures them through files on disk, so they can't
 carry a key. `llmman serve` needs the variable in its own environment
-instead. That
-fallback is only used for a daemon bound to loopback, and never for a
+instead. That fallback is only used for a daemon bound to loopback, and never for a
 browser request from another site. It bounds the blast radius rather
 than authenticating anyone, so on a shared machine prefer an integration
 that sends its own key. `cline`, `kimi`, `copilot` and `openclaw` can't be
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ running on your own machine, or at any hosted provider, in one command.
 No subscription, no rewiring, no vendor's idea of which model you should use.
 </p>
 
-```
+```sh
 llmman launch claude --model gemma4
 ```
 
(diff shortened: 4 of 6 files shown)


ericcurtin:
Missing e2e tests

ericcurtin (commented):
Config merging and the tests are good. Four things first.

- Selection rests entirely on `droid --model custom:llmman-0`. Set `sessionDefaultSettings.model` to that id in `settings.json` instead, plus `reasoningEffort` when the existing value is missing or invalid, so it does not depend on the flag accepting a custom id. Was that flag actually run against a real `droid`? The verification list only shows a registry listing.
- `entry.insert("apiKey", "llmman")` hardcodes the key. Use `providers::PLACEHOLDER_API_KEY`.
- `supportsImages` is never written, so a vision model silently loses images. Set it from the model's capability.
- README: the paragraph is rewrapped badly ("instead. That" mid-line) and the trailing newline at EOF is dropped, both unrelated to this change.

ericcurtin:
There are conflicts to fix

ericcurtin (commented):
Conflicts with main. `PROVIDER_NEEDS_DAEMON_KEY`, `INTEGRATIONS` and the README hosted-provider paragraph all moved under #355 and #374. Rebase on main and re-resolve those three spots, then re-run the unit tests.

On the implementation:

1. Model selection should go through `sessionDefaultSettings.model` in the settings file, not a `--model custom:llmman-N` argv flag. That is the documented selection surface and it survives a session the user starts by hand. Set `reasoningEffort` to a valid value at the same time, or Droid falls back to whatever stale value is in the file.
2. `llmmanManaged` is a field Factory's schema does not define. Do not inject one. The entry already carries `apiKey: "llmman"`, which is enough to recognise ownership, and it is what the equivalent integration elsewhere keys on.
3. Hardcoded `"llmman"` for the key should be `providers::PLACEHOLDER_API_KEY`.
4. `supportsImages` is never set, so a vision model loses image input in Droid.
5. The write replaces the user's settings with no backup. Write a `.bak` first.
6. The README diff drops the trailing newline at end of file. Unrelated, revert it.

ericcurtin:
Converting this to draft until it's ready for review again

ericcurtin:
@mah1104ahm are you planning on continuing with this one?

mah1104ahm:
Yes, I am continuing, and I have now pushed the follow-up in b304098.
Thank you for the detailed review and for checking back.

The branch is rebased onto main with the conflicts resolved. Droid now
selects the model through sessionDefaultSettings, preserves valid
reasoning settings, identifies the managed entry by the placeholder
API key, removes llmmanManaged, and sets image flags from available
model metadata. Settings are backed up and replaced atomically, with
cooperating launches serialized. The README wrapping and final newline
are corrected.

Validation: all 20 launcher unit tests passed. I also installed Droid
0.208.1 and the CI-pinned llama.cpp b10293, then ran the new real CLI
E2E test against qwen3.5:0.8b on macOS arm64. The model warm-up and
the actual llmman launch droid command both succeeded; the test
verified the expected pong response without a --model override to
Droid. The pinned CLI install and E2E test are included in the
workflow.

The PR remains draft while the new upstream checks run. This is
verified locally on macOS arm64; I am not claiming the other platform
runs have passed yet.

M Ahmed

mah1104ahm:
I pushed follow-up e56b1a0 addressing the latest review: managed
entries now require the expected llmman display name, ID prefix and
numeric index in addition to the placeholder key; unknown capabilities
clear both image flags; and the real Droid E2E uses the suite's
existing retry/tolerant response helper.

Validation on macOS arm64: formatting and whitespace checks passed,
all 22 launcher unit tests passed, and the real Droid 0.208.1 E2E
passed with llama.cpp b10293 and qwen3.5:0.8b. The shared response
helper tolerates sampling variance, so I am not claiming strict
response-content validation or cross-platform success.

The PR description now reflects this evidence. GitHub reports
action_required for the new CI run, with no check runs yet:
https://github.com/llmmanorg/llmman/actions/runs/33758531185 . Could a
maintainer check whether workflow approval is needed? The PR remains
draft pending upstream checks and review.

M Ahmed

ericcurtin (changes requested):
Selection through `sessionDefaultSettings` and the backup are right now. Conflicts first, then five things.

The README hosted-provider paragraph moved again under #396, which added provider keys in `llmman.conf`. Rebase on main and re-resolve `README.md`: keep main's new `llmman.conf` text and only add `droid` next to `hermes` in the sentence that names the exception.

1. `noImageSupport` is not a field Factory's schema defines. Same objection as `llmmanManaged` earlier. Drop it; `supportsImages` is the documented flag.
2. `displayName: "llmman"` makes every entry read as "llmman" in Droid's picker. The reference uses the model name there. Use the model name.
3. `is_llmman_droid_entry` now demands four fields. The reference recognises its own entry by the placeholder key alone, and the closest analogue in this repo uses key plus base URL. With four, a user renaming the entry orphans it and the next launch appends a duplicate. Match on `apiKey == PLACEHOLDER_API_KEY` and `baseUrl`.
4. `droid_image_support` opens the OCI store straight from the launch process. The daemon already owns model metadata and this path returns `None` for every `--provider` launch. Ask the daemon for the capability instead.
5. `tempfile` is a new dependency for one atomic write. #379 adds `fsutil::write_atomic` for exactly this. Reuse it and drop the dep.

Nits: the trailing newline at EOF in README is still dropped, and the ```` ```sh ````/```` ```powershell ```` fence relabelling is unrelated to Droid. Empty `--model` persists a custom model literally named `default` into the user's settings; the reference writes nothing when there is no model.

ericcurtin:
Converting to draft. Mark it ready for review again when the rebase and the above are done and I'll take another pass.

PR #353, opened 2026-09-01
fix: match ollama's unload contract on the API and in llmman stop

ricky-chaoju:
Follow-up to #329.

`unload_key` now returns a provider-routed reference untouched, the one step it still did not share with `ensure_model`. The unload handlers route through a new `unload_model`, which drops the running entry if there is one and otherwise consults the store: a model llmman has no record of gets a 404, a model it holds but has not loaded gets the usual success. That is ollama's contract (checked against 0.32.6), and running is consulted first so a model removed from the store while still loaded stays unloadable. `llmman stop` sends the unload directly and turns the 404 into its error, which removes the ps round trip and the window between the two calls, and it prints nothing on success like `ollama stop`.

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -2287,10 +2287,20 @@ fn unload_key(
 /// unloadable, since the `running` entry is authoritative and is consulted
 /// first.
 async fn unload_model(state: &AppState, model: &str) -> Result<(), AppError> {
-    let canonical = unload_key(state, model).map_err(AppError::bad_request)?;
-    // Wait for an in-flight load of this model to publish itself first, so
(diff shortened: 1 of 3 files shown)


ericcurtin:
Merged. Verified locally with #338, #354 and #355 also applied: 443 unit tests pass, clippy and fmt clean.

Follow-ons:

- The load lock key is now pre-canonical, so `m@sha256:...` and `m:latest` for the same stored model take two different locks and can both load. Narrow window, but it used to be one lock.
- `daemon::unload` reads any 404 as "no such model", including one from a proxy in front of the daemon or a wrong path. Checking the body keeps the message honest.

PR #354, opened 2026-09-01
refactor(shortnames): parse a reference once per resolve

maxgio92:
Follows up on PR #303: `resolve` and `resolve_ollama_api` each parsed a registry-grammar reference twice.
`validate_reference` calls `parse_registry_ref` once, then `resolve_inner`'s `has_host` call reparsed the same string.

This threads the already-parsed `ParsedRef` into `resolve_inner` instead of reparsing.
`has_host` is deleted; its two unit tests now assert the same behavior through `resolve`'s output.

No behavior change: full test suite passes unmodified, and the diff is scoped to _src/shortnames.rs_ only.

### DIFF
--- a/src/shortnames.rs
+++ b/src/shortnames.rs
@@ -361,6 +361,15 @@ fn rejoin(parsed: &ParsedRef<'_>) -> String {
 ///       malformed input fails by construction rather than by enumerating
 ///       bad shapes.
 pub fn validate_reference(reference: &str) -> Result<(), InvalidReference> {
+    validate_reference_parsed(reference).map(|_| ())
+}
+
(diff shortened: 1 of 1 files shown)


ericcurtin:
Merged. Confirmed behaviour-preserving: `parse_registry_ref` handles the scheme itself, so the host check on the pre-strip parse matches the old post-strip one.

Follow-on: `has_host` is gone, so rename `has_host_requires_a_slash` and `has_host_recognizes_an_explicit_port` after what they now assert about `resolve`.

maxgio92:
Thanks Eric. Going to follow-up on a dedicated PR

PR #355, opened 2026-09-01
feat(launch): add Qwen Code integration

### LINKED ISSUE #352: Add "llmman launch qwen"
And test aggressively to ensure it works before we open the PR.

ricky-chaoju:
Qwen Code has an OpenAI-compatible mode, so this points `OPENAI_BASE_URL` at the daemon's `/v1`, passes the key in `OPENAI_API_KEY`, and execs. Auth type and model also go on the command line as `--auth-type openai --model <model>`, because Qwen Code reads its arguments, then `~/.qwen/settings.json`, then the environment, and writes that file itself when a user picks a provider with `/auth` or `/model`. With such a file present, a launch through the variables alone went to the cloud provider it named, with that provider's key, and reported success.

Ollama's `cmd/launch/qwen.go` prepends the same two flags. Each yields to the caller's own after `--`, since a repeated flag crashes Qwen Code 0.22.3. The key stays in the environment because on the command line it shows in `ps`. `--model` is required: with none, Qwen Code sends its built-in default `qwen3.7-max`, which the daemon would then try to pull and fail on, so the launcher refuses first and names the flag.

Closes #352.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -422,7 +422,7 @@ jobs:
         run: go test -tags podman,containers_image_openpgp,exclude_graphdriver_btrfs,exclude_graphdriver_devicemapper ./...
 
   # --------------------------------------------------------------------------
-  # E2E: `llmman launch <claude|opencode|codex|hermes|openclaw> --model
+  # E2E: `llmman launch <claude|opencode|codex|qwen|hermes|openclaw> --model
   # qwen3.5:0.8b` (tests/launch_e2e.rs) against a real `llmman serve`, a
   # real pulled model, a real llama-server, and the real third-party CLIs
   # — not mocks. Mirrors the `build` job's own 5-target matrix exactly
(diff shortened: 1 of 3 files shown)


ericcurtin:
Merged. The flag/env precedence and the `--auth-type openai --model` prefix match how this is done elsewhere.

Follow-ons:

- Also write `~/.qwen/settings.json`, the way `write_codex_config` and `write_hermes_config` do for theirs: an `openai` provider entry (id, baseUrl, envKey), `security.auth.selectedType`, `security.auth.baseUrl`, and `model.name`. Merge in place so unrelated settings survive. Right now a persisted `modelProviders` entry whose id equals the launched model still beats argv and env, which is the failure the doc comment describes.
- `find_on_path("qwen")` misses npm, nvm, homebrew and `%APPDATA%\npm` installs. Same candidate-path fallback `launch_opencode` already has would cover it.

ericcurtin:
Squashed onto main as c1da3c1. Merged from the command line rather than the button because the PR touches `.github/workflows/ci.yml`, so GitHub will not mark this PR merged; closing it by hand.

ricky-chaoju:
Curious what blocked the merge button here. CI was green and it merged cleanly, and the earlier ci.yml PRs went through the button fine.

PR #356, opened 2026-09-01
fix(ci): pin vllm-metal's stable release for the macOS e2e

ricky-chaoju:
The macOS e2e job has failed on every push since #338: the pinned `vllm_metal-0.3.0.dev20260825040033` wheel 404s, since vllm-metal now deletes each dev prerelease when the next one is published (vllm-project/vllm-metal#662). This pins the stable `v0.28.0` instead, and moves `VLLM_VERSION` to the 0.28.0 it is built against; the Linux CPU legs take that bump too. The Linux aarch64 leg passes `pytest -m e2e -v` locally with vLLM 0.28.0; macOS is left to CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,13 +27,18 @@ env:
   # independently.
   LLAMA_CPP_RELEASE: b10293
   # vLLM version the e2e job's "Install vLLM (e2e)" step installs, for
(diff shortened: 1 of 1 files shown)


ricky-chaoju:
@ericcurtin this is the fix for the red CI on main.

ericcurtin (approved):
Thank you @ricky-chaoju !

PR #357, opened 2026-09-01
Remove unsupported vLLM plugin store override

LxYuan0420:
This PR is:

- To remove a stale `store=` wrapper option that forwarded an unsupported `--store` flag.
- To align the vLLM plugin resolver with llmman’s `LLMMAN_MODELS` store override.
- To delete obsolete `--store` test coverage while retaining `--cache` coverage.

The wrapper accepted a private `store=` argument and translated it into a CLI flag that does not exist:

```text
llmman resolve <reference> --store <path>
```

That fails with:

```text
error: unexpected argument '--store' found
```

The normal vLLM plugin path does not pass `store=`, so this is a small contract cleanup rather than a broad runtime bug.

`llmman resolve` supports `--cache`, but not `--store`; the store override is `LLMMAN_MODELS`.

### DIFF
--- a/vllm-plugin/src/vllm_llmman/_llmman_cli.py
+++ b/vllm-plugin/src/vllm_llmman/_llmman_cli.py
@@ -39,7 +39,6 @@ def llmman_binary() -> str:
 def resolve(
     reference: str,
     *,
-    store: str | None = None,
     cache: str | None = None,
     binary: str | None = None,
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):

ericcurtin:
Thanks @LxYuan0420

PR #369, opened 2026-09-01
feat(serve): expose a Prometheus metrics endpoint

### LINKED ISSUE #359: serve: expose a /metrics endpoint
`llmman serve` exposes no metrics endpoint today. There are zero hits for
`prometheus` or `/metrics` anywhere in `src/`, so operators running llmman as a
daemon have no way to scrape request counts, queue depth, load/unload events, or
per-model latency.

Scope: a `/metrics` route on the serve router, Prometheus text format, covering
at minimum request count and duration per endpoint, currently loaded models, and
model load/unload counts.

@anagnorisis2peripeteia, you added a metrics endpoint to Ollama in
https://github.com/ollama/ollama/pull/16998, which is the same shape as what
this needs. No obligation at all, but if you want to take this, or just tell us
what you would do differently the second time around, that would be useful.

anagnorisis2peripeteia:
## What

Adds `GET /metrics` — Prometheus text exposition format, **off unless `LLMMAN_METRICS=1`** — and the module behind it, `src/metrics.rs`.

Fifteen metric families:

| Metric | Type | Labels |
| --- | --- | --- |
| `llmman_build_info` | gauge | `version` |
| `llmman_start_time_seconds` | gauge | — |
| `llmman_scheduling_requests_in_flight` | gauge | — |
| `llmman_scheduling_capacity` | gauge | — |
| `llmman_scheduling_rejections_total` | counter | — |
| `llmman_models_loaded` | gauge | — |
| `llmman_models_loading` | gauge | — |
| `llmman_model_up` | gauge | `model`, `engine` |
| `llmman_model_loads_total` | counter | `model` |
| `llmman_model_load_duration_seconds` | histogram | `model` |
| `llmman_model_load_oom_retries_total` | counter | `model`, `strategy` |
| `llmman_model_unloads_total` | counter | `model`, `reason` |
| `llmman_http_requests_total` | counter | `route`, `status` |
| `llmman_http_request_ttfb_seconds` | histogram | `route` |
| `llmman_http_request_duration_seconds` | histogram | `route` |

`reason` is `idle`, `requested`, `crashed`, `oom` or `evicted`, one per site that removes an entry from `running`. `strategy` is `evict_others`, `split_mode` or `ctx_shrink`, one per OOM fallback `ensure_model` tries. Both are closed enums; `model`, `route` and `version` are not.

Two behaviour changes outside the new module:

- `reap_idle_models_once` checks `is_alive` before labelling an unload, and its log line follows the same branch. A model that died and was then never asked for again reaches its `keep_alive` deadline looking exactly like a healthy idle one, so without the check `crashed` would read zero on the daemon whose backends are dying.
- `PendingLoadGuard` releases its reservation *into* the manager lock that inserts into `running`, and the unbounded `LLMMAN_MAX_LOADED_MODELS` path now reserves as well. Previously the guard released asynchronously, after `ensure_model` had already inserted, so a load was counted twice for as long as that task took to run. That window is what made `llmman_models_loading` unimplementable before; closing it is what makes `loaded + loading` equal exactly what `LLMMAN_MAX_LOADED_MODELS` caps.

## Why

`llmman serve` published nothing an operator could scrape. There are no hits for `prometheus` or `/metrics` anywhere in `src/`, so request rate, scheduling depth and model lifecycle were visible only in log lines.

The numbers worth having are the ones no backend can publish, because they are about the router rather than about inference:

- **Cold start.** `llmman_model_load_duration_seconds` times from llmman finding a model unloaded to it answering — queue wait, pull, eviction and spawn included. For a daemon that idle-reaps and LRU-evicts, cold start is the tail latency operators actually feel, and llama-server cannot see it because it is not running yet.
- **Which model.** `llmman_model_unloads_total{reason="evicted"}` says eviction is happening; the `model` label says which one is thrashing, which is the part you act on. Same for load latency and OOM retries.
- **Saturation.** `llmman_scheduling_rejections_total` counts the 503s from `MAX_QUEUE_ERROR`. A saturated daemon was otherwise invisible outside the log.
- **Liveness.** `llmman_model_up` is 1 while a loaded model's backend is alive and 0 once it has exited. `check_running` only inspects a process when a request arrives for that model, so `sum by (instance) (llmman_models_loaded) - sum by (instance) (llmman_model_up)` is the count of dead backends llmman has not noticed yet.

## Notes

- **Off by default.** The router has no authentication, `LLMMAN_HOST` can bind it beyond loopback, and a scrape reports version, route mix, model names and model churn. None of that should start answering because llmman was upgraded. Unset, the route is not registered at all and answers `404` rather than `403`, so a disabled endpoint is indistinguishable from a build that never had one.
- **Two histograms, not one.** `ttfb` stops at response headers; `duration` stops when the body ends. On `/api/version` they agree. On `/api/chat` with `"stream":true` they do not, and only `duration` includes generation. A single family named `request_duration_seconds` that stopped at headers would mislead whoever graphed it. The end-of-body hook is the same one `proxy` already uses: an RAII value moved into the stream closure, dropped when the body finishes.
- **The `Content-Length` trap.** hyper derives `Content-Length` from the body's `size_hint()`, and a body wrapped in `Body::from_stream` no longer has an exact one. Timing every body the same way therefore turned every JSON reply on the daemon into `transfer-encoding: chunked` — measured against a real daemon as `content-length: 84` becoming chunked on `/api/version`. `track_metrics` now wraps only bodies whose size is not already known; those are the streaming ones, which were chunked either way. A body already complete when the headers go out has no generation left to time, so the second clock would have read the same as the first.
- **Cardinality.** `model` is bounded by what an operator has actually asked this daemon to run, `route` by axum's `MatchedPath` rather than by traffic. An unmatched request carries no `MatchedPath` and is not counted; neither are CORS preflights, which the outer layer answers first. Every per-model family is written on a load, unload, OOM retry or scrape of `running`, so nothing grows from request volume.
- **Label escaping is load-bearing.** `route` and the two enums could not need it. `model` can: `validate_reference`'s absolute-path branch (`shortnames.rs:385`) rejects only ASCII control characters, so `/models/a"b\c.gguf` is a valid model reference, and both of those characters have to be escaped for the exposition to parse. A carriage return is dropped rather than escaped — the format defines exactly three sequences (`\\`, `\"`, `\n`), so an escaped `\r` fails the parse of the whole exposition rather than of its own line, and `promtool check metrics` rejects it with `invalid escape sequence '\r'`. Left raw it parses but puts a bare control character in a line-oriented format.
- **No new dependency.** A `prometheus` client crate brings a registry, a descriptor system and a protobuf encoder to emit ~150 lines of text. The module is `format!` into a `String` behind one `Mutex<HashMap>`; the `HttpBody` trait needed for the size-hint check is axum's own re-export.
- **`/metrics` is the one route outside the CORS layer**, so it answers without an `Access-Control-Allow-Origin` header. `default_allowed_origins` is appended unconditionally, so any page on any localhost port is an allowed origin. Nothing in a browser scrapes a metrics endpoint, so leaving the layer off costs nothing. What makes that true is only the order of `Router::merge` against `Router::layer`, which is easy to undo by accident, so a test asserts the header's absence against the real router.
- **No per-token counters.** llmman does not run inference; llama-server does, and already publishes prompt/eval totals and KV-cache usage on its own `/metrics` behind `--metrics`. A copy here would mean either scraping every backend on every scrape, or numbers that drift from the engine's own. Passing `--metrics` through to the backend spawn is a separate change.
- **The scrape holds the manager lock** for one pass over `running`, the same lock `/api/ps` already takes. Nothing on the load path holds it across slow work: the pull, all four spawn paths, `wait_for_ready` and `stop_and_wait` all run with it dropped.
- **Zero-fill policy.** An unlabelled counter always emits a sample, so `llmman_scheduling_rejections_total` reads 0 on a healthy daemon rather than vanishing. A labelled family emits nothing until it has an observation, so no label value is invented before it exists.
- **Disabled means uninstrumented.** The `track_metrics` layer is applied only when `LLMMAN_METRICS` is set, not just the route. Enabling the endpoint needs a restart, so a disabled daemon that kept recording would pay a registry lock and a map write per request, and grow a `model` label set for the life of the process, for numbers nothing could scrape.

## Test

`cargo test --lib --bins` (469 + 13), `cargo clippy --release --all-targets -- -D warnings` and `cargo fmt --all --check`, clean. Only `aarch64-apple-darwin` locally; the other four targets are CI's.

`cargo mutants --in-diff` over this PR's diff: 83 mutants, 3 survivors. Two of those mutate `metrics_enabled_from_env`, which follows the repo's `X_from_env` / `parse_X` split where only `parse_X` is tested; the third replaces the whole of `serve_async`, which binds a listener and serves forever. The three it caught that mattered are in the third commit; the fourth commit trims the tests to this repo's density and re-ran the gate to prove nothing it cut was load-bearing.

Against a daemon serving `qwen3.5:0.8b` through llama-server:

- `/metrics` answers `404` with no `LLMMAN_METRICS`, `200` with `LLMMAN_METRICS=1`, while `/api/version` answers `200` either way, and a request through the disabled router leaves no series in the registry.
- On `/api/chat` with `"stream":true`: TTFB 2.013s, total duration 2.725s, same request. The two histograms disagree by the generation time, which is the point of splitting them.
- `content-length` present and `transfer-encoding` absent on every buffered route, `/metrics` included.
- `llmman_models_loading` reached 1 during a cold start and returned to 0, with `llmman_models_loaded` never double-counting the same load.
- `llmman_model_up` read 1, then 0 after the backend was killed, while `llmman_models_loaded` still counted the corpse.
- `requested`, `evicted`, `idle` and `crashed` unloads were each provoked against real backends, labelled by model. `oom` is covered by a test driving `evict_other_models` directly, since `LLMMAN_CONTEXT_LENGTH` is daemon-wide and any value large enough to fail a second load fails the first.
- `llmman_scheduling_rejections_total` read exactly 39 against 39 observed 503s.
- `llmman_scheduling_requests_in_flight` reached 5 under five concurrent cold starts.
- 2001 requests through the middleware, request counter and histogram count exact. 10,000 concurrent recordings over five rounds, counter exact each time.
- `Origin: http://localhost:3000` gets `access-control-allow-origin` back from `/api/version` and nothing from `/metrics`. The regression test fails on the intended message when `/metrics` is moved above the layer.

Prometheus 3.14.0 scraped the endpoint on its default path, no `metrics_path` line. `rate()`, `histogram_quantile()`, `rate(_sum)/rate(_count)` and a `group_left` join onto `llmman_build_info` all return values; so does `sum by (instance) (llmman_models_loaded) - sum by (instance) (llmman_model_up)`. `promtool check metrics` is clean on both a cold and a populated exposition.

Closes #359.


ericcurtin (commented):
Apologies. I've had a think and propose a redesign.

The instrumentation points you chose are the right ones. This is about what gets published and who can see it, not about the plumbing.

## 1. Off by default, `LLMMAN_METRICS=1` to enable

The one change I'd insist on. `/metrics` should not answer unless the operator asked for it.

The reason is exposure. The router has no authentication at all (`serve.rs:2800`), `LLMMAN_HOST` will bind it anywhere, and the endpoint currently sits inside `cors_layer`, so a web page can read it.

## 2. Path: `/metrics`, and drop `/llmman/metrics`

Good news, lets promote this to `/metrics`

## 3. The missing metric that matters most: cold-start latency

`llmman_model_loads_total` counts loads and never times them. Load latency is the most valuable number this daemon can publish, and the only one no backend can ever measure. For a router that idle-reaps (`reap_idle_models_once`, `serve.rs:1715`) and LRU-evicts (`enforce_max_loaded_models`, `serve.rs:2417`), cold start *is* the tail latency operators actually feel.

## 4. Add a `model` dimension

Nine families and not one carries a model label. Bounding cardinality was the right instinct, and I'd have raised it if you hadn't, but you bought it by dropping the one dimension llmman uniquely has. `llmman_model_unloads_total{reason="evicted"}` tells me eviction is happening; it can't tell me *which* model is thrashing, which is the part I'd act on. Same for load latency.

## 5. Duration: split TTFB from stream duration

You flagged this yourself in the notes. On the streaming routes, stopping the clock at response headers measures time to first byte, and naming it `request_duration_seconds` will mislead whoever graphs it. TTFB is worth having, it's the nearest thing we can offer to TTFT but name it that. Total duration is reachable too: `proxy` already holds the `ActivityGuard` inside the stream closure until the body finishes (`serve.rs:3253`), so the end of the stream has a hook.

## 6. Also missing

- **Queue rejections.** `MAX_QUEUE_ERROR` (`serve.rs:2482`) is a 503 that needs a counter. Today a saturated daemon is invisible outside the log.
- **OOM retries.** The retry loop around the spawn is a direct signal about undersized hardware.
- **Backend liveness**, per model and engine. The reaper now separates crashed from idle, which is good, but nothing says a backend is down *right now*.

On `pending_loads`: your reasoning for leaving it out is sound, `PendingLoadGuard` releases after `ensure_model` has already inserted into `running`, so a naive gauge double-counts. But #359 asks for the loading count, so the answer is to fix the release ordering, not to document the gap.

## 8. On size

1260 additions for nine families, with a hand-rolled registry, label escaping and golden exposition tests, is more than this needs, and the redesign above adds metrics rather than lines. Before the next round, check how much of `src/metrics.rs` survives contact with `format!` into a `String` behind one `Mutex<HashMap>`. If the escaping helper only ever sees route names from `MatchedPath` and a closed `reason` enum, it may not need to exist at all.

anagnorisis2peripeteia:
Redesign done, all of it. Five commits on top of the one you reviewed rather than an amend, plus a merge from main, so your "changes since your review" diff stays readable; they squash away at merge.

Surface went from nine families to fifteen. Taking your numbers in order.

## 1. Off by default, `LLMMAN_METRICS=1`

Done. `parse_metrics_enabled` accepts `1`, `true`, `yes`, `on`, case-insensitively, and nothing else.

Unset, the route is **not registered**, so it answers `404` rather than `403` — a disabled endpoint is indistinguishable from a build that never had one.

Off also means off rather than hidden: the `track_metrics` layer is applied only when the flag is set. Enabling the endpoint needs a restart, so a disabled daemon that kept instrumenting would pay a registry lock and a map write per request, and grow a `model` label set for the life of the process, for numbers nothing could ever scrape. I had this wrong until the mutation run below; the first version hid the endpoint and measured anyway.

`build_router` takes the flag as an argument rather than reading the environment, so the off case is testable against a real router. `the_scrape_endpoint_is_absent_unless_the_operator_enabled_it` asserts `404` on `/metrics`, `200` on `/api/version`, and that a request through the disabled router leaves no series behind.

You were right about the exposure. I had the CORS reasoning and stopped there.

## 2. Path

Done, in `dd0708a`. `/llmman/metrics` is gone, and the README scrape config loses its `metrics_path` line.

That commit also moved the routes into `build_router` so `/metrics` sits outside `cors_layer` — `Router::layer` applies only to routes registered before it, which is an ordering easy to undo by accident, so a test asserts the header's absence against the real router rather than against a copy of the layering. Same commit fixes `escape_label_value` to drop a carriage return rather than escape it: the format defines exactly three sequences, so an escaped `\r` fails the parse of the whole exposition, and `promtool` rejects it with `invalid escape sequence '\r'`.

## 3. Cold-start latency

`llmman_model_load_duration_seconds{model}`, histogram, same bucket ladder as the HTTP families.

The clock starts where `ensure_model` finds the model unloaded and stops when it is in `running` and answering — so queue wait, pull, eviction and spawn are all inside it, which is what the caller actually waited for. Not the spawn alone.

`llmman_model_load_oom_retries_total{model,strategy}` sits beside it, `strategy` being `evict_others`, `split_mode` or `ctx_shrink` — one per fallback the retry loop tries, so a rising `ctx_shrink` says the box cannot hold the context this daemon picked, and `evict_others` says it cannot hold two models at once. Different hardware conversations.

## 4. `model` dimension

On every per-model family: loads, load duration, OOM retries, unloads, liveness. `llmman_model_unloads_total{model="…",reason="evicted"}` now names the thrashing model.

Cardinality is bounded by what an operator has asked this daemon to run, and every one of those families is written on a load, unload, retry or scrape of `running` — none of them grow with request volume.

One consequence worth flagging, because it bears on 8: **`model` is what makes `escape_label_value` load-bearing.** `validate_reference`'s absolute-path branch (`shortnames.rs:385`) rejects only ASCII control characters, so `/models/a"b\c.gguf` is a valid model reference today, and `"` and `\` both have to be escaped or the exposition does not parse. Routes and the two enums could not have needed it; model names can.

## 5. TTFB and total duration

Split.

- `llmman_http_request_ttfb_seconds{route}` — stops at response headers.
- `llmman_http_request_duration_seconds{route}` — stops when the body ends.

Your pointer to `ActivityGuard` was the right one. `track_metrics` moves an RAII timer into the stream closure so it drops with the body.

One thing that only showed up against a live daemon: hyper derives `Content-Length` from the body's `size_hint()`, and a body wrapped in `Body::from_stream` no longer has an exact one. Wrapping every body turned every JSON reply on the daemon into `transfer-encoding: chunked` — `content-length: 84` on `/api/version` became chunked, to measure a body that was already complete when the headers went out. So the wrapper is now behind a `size_hint().exact().is_none()` check: buffered bodies record at the headers, where the second clock would read the same anyway, and only genuinely streaming bodies get wrapped. Those were chunked either way. `timing_a_response_does_not_cost_it_its_content_length` guards it against a live response, since the header is not one this middleware sets.

Measured on `/api/chat` with `"stream":true`: TTFB 2.013s, total 2.725s, same request.

## 6. The three missing signals

- **Queue rejections.** `llmman_scheduling_rejections_total`, incremented where `try_admit` returns `MAX_QUEUE_ERROR`. Unlabelled, so it zero-fills: a healthy daemon publishes `0` rather than nothing. Read 39 against 39 observed 503s.
- **OOM retries.** Covered in 3.
- **Backend liveness.** `llmman_model_up{model,engine}` — 1 while the backend process is alive, 0 once it has exited. Because `check_running` only inspects a process when a request arrives for that model, `sum by (instance) (llmman_models_loaded) - sum by (instance) (llmman_model_up)` is the count of dead backends llmman has not noticed yet. Verified by killing a backend: `model_up` went to 0 while `models_loaded` still counted the corpse.

On `pending_loads`: you were right, and I fixed the ordering. `PendingLoadGuard` now releases its reservation *into* the same manager lock that inserts into `running`, so there is no window where a load is in both; and the unbounded `LLMMAN_MAX_LOADED_MODELS` path reserves too, which it previously did not. `llmman_models_loading` is the reservation count, and `loaded + loading` is now exactly what `LLMMAN_MAX_LOADED_MODELS` caps. Two tests cover the closed window and the unbounded path.

## 8. Size

Straight answer first: **the PR got bigger, not smaller.** 1,260 additions when you costed it, 1,838 now. I am not going to dress that up.

The `format!`-into-a-`String`-behind-one-`Mutex<HashMap>` design is what it already was — no registry, no descriptors, no protobuf, no client crate. There was no rewrite available; that is the shape.

So the only honest measurement is per family, and by that it did get denser:

| | `6964c9e` | now |
| --- | --- | --- |
| families | 9 | 15 |
| `metrics.rs` production lines | 470 | 667 |
| per family | 52.2 | **44.5** |
| code lines only, no comments or blanks | 262 | 393 |
| per family | 29.1 | **26.2** |

Where the other 578 lines went: 439 production, 128 tests, 11 README. Over half of the new production is `serve.rs` call sites rather than the module — the cold-start timer, the release-ordering fix, the liveness scrape, the body wrapper — which is where six families' worth of instrumentation actually lives.

On the escaping helper: it stays, for the reason in 4. It is 13 lines and it now has an input that is not a closed set.

It was 2,327 before the last commit. That one cut the tests to this repo's own density — the byte-for-byte golden exposition, a fuzz harness guarding the 13-line escaping helper, and test docs written at production density all went; nothing else here does any of those. The mutation run below was re-run after the cut and caught the one place it had gone too far, so what is left is the set that does work.

## The bot threads

Both of the original ones are closed by the above, so I have not replied on them separately. CodeRabbit's three README notes on the new pushes — the `LLMMAN_HOST` wording overstated the default bind, and it was right — are the fifth commit.

CodeRabbit wanted `llmman_models_loading` and proposed rendering `pending_loads` directly. That gauge now exists, but not that way: reading `pending_loads` as it stood would have counted a load twice for as long as the guard's release task took to run, since the guard released after `ensure_model` had already inserted into `running`. The fix is the release ordering; the gauge falls out of it.

Copilot flagged that `evict_other_models` increments the process-wide `oom` counter from a parallel test without the lock its exact-delta assertion depends on. That test holds the lock as of `dd0708a`, and the `model` label has since removed the collision from the other side too: the metrics-side test asserts against a `wrapper-probe` model no other test writes. The one unlabelled counter it still asserts a delta on is `llmman_scheduling_rejections_total`, and no test reaches the production increment — the queue tests go through `try_admit_against` with their own static, which deliberately does not count.

## What the third and fourth commits are

`cargo mutants --in-diff` over this PR's own diff, 83 mutants. Six survived. Three were real and are fixed:

- `record_oom_retry` was the one wrapper the shared-registry test never called — exactly the failure that test exists to catch, since a wrapper that drops its increment leaves a live daemon reporting a flat counter while every `*_into` test keeps passing.
- `RunningModel::engine_label` had no test at all, so the `engine` label on `llmman_model_up` could be emptied or renamed silently.
- The `track_metrics` gate in 1.

The other three I left. Two mutate `metrics_enabled_from_env`, which follows this repo's `X_from_env` / `parse_X` split where only `parse_X` is tested — eight of the eleven `_from_env` helpers in `serve.rs` have no direct test, and reaching them means setting process-wide environment variables inside a parallel test binary. The third replaces the whole of `serve_async`, which binds a listener and serves forever.

The fourth commit is the size cut in 8. The gate re-run after it found two new survivors, both deleted arms of `escape_label_value` that only the removed fuzz harness had been catching; one five-line direct test of the helper replaced it and the gate is back to the same three.

---

`cargo test --lib --bins` 469 + 13, clippy `-D warnings` and `fmt --check` clean. Prometheus 3.14.0 scrapes it; `rate()`, `histogram_quantile()`, `rate(_sum)/rate(_count)` and a `group_left` join onto `build_info` all return values. `promtool check metrics` clean on cold and populated expositions. README rewritten with the full table and the `LLMMAN_METRICS` row.

ericcurtin:
Merged. Verified locally: suite, clippy and fmt clean, `/metrics` 200 with `LLMMAN_METRICS=1` and 404 without, no `access-control-allow-origin` on the scrape, `content-length` intact on buffered routes, both histograms counting.

Follow-ons, separate PRs:

1. The unbounded `LLMMAN_MAX_LOADED_MODELS` path now arms `PendingLoadGuard`, whose `Drop` spawns the decrement. A decrement lost to runtime shutdown drifts `llmman_models_loading` up for the life of the process. Make the release synchronous on the failure paths too.
2. `escape_label_value` silently drops `\r`. Reject the reference at validation instead, so the store never holds a name that cannot round-trip.
3. Pass `--metrics` through to the llama-server spawn so the backend's own counters are scrapable next to these.
4. No e2e coverage. A scrape after a real load in `launch_e2e.rs` would catch a family that stops being written.

PR #376, opened 2026-09-02
 fix: resolve a digest reference to the stored model

ricky-chaoju:
Follow-up to #353.

The first follow-on there described `m@sha256:…` and `m:latest` taking two load locks. Reproducing it found the store underneath: `ref_matches_precise` compared names only, so a digest reference never matched a stored model. `find` said "not found", and the daemon pulled the same bytes again under a second reference and started a second `llama-server` for them, no race involved. It now matches `<name>[:tag]@<digest>` by manifest digest under the same repository, whatever tag holds it. That changes what `rm`, `/api/delete`, `show`, `canonical_ref`, and the pull decision resolve a digest to; with several tags at one digest, the spelled tag wins, and `rm` takes one entry per call.

`load_identity` folds a `@digest` onto the tag spelling, and `ensure_model` now takes its lock key from it too, so the digest and `:latest` spellings share one lock while the reference the store is asked about keeps the digest. What remains is a first load of `m:v9` and one of its digest arriving together, since which tag a digest sits under is not known before the store is read.

`daemon::unload` read any 404 as "no such model". It now checks for the `{"error":"model '…' not found"}` envelope and reports any other 404 with its status and body.

### DIFF
--- a/src/cmd/cp.rs
+++ b/src/cmd/cp.rs
@@ -33,14 +33,41 @@ pub fn run(args: &CpArgs) -> anyhow::Result<()> {
     let store = OciStore::open(&store_root)?;
 
     let desc = store.find(&source)?;
+    // A destination spelled as a digest names content, and the store
+    // answers such a reference by content (see `OciStore::find`): a
+    // pointer there holding some other digest would be passed over by a
+    // lookup for the digest it is named for and answer only for the one
+    // it holds, so it is refused.
+    check_digest_destination(&args.destination, &desc.digest)?;
     store.tag(desc, &args.destination)?;
     println!("copied '{}' to '{}'", args.source, args.destination);
     Ok(())
 }
 
+/// Refuses a `<name>@<digest>` destination whose digest is not `actual`,
+/// the copied descriptor's. A tag, or the matching digest in either
+/// case, passes.
+fn check_digest_destination(destination: &str, actual: &str) -> anyhow::Result<()> {
+    match crate::storage::split_ref_digest(destination).1 {
+        Some(digest) if !digest.eq_ignore_ascii_case(actual) => anyhow::bail!(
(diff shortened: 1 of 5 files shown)


ericcurtin (commented):
Good root-cause find, and the tests cover the interesting cases. Two things before merge:

1. `rm` by digest is now silently ambiguous and destructive. With `m:v8` and `m:v9` at one digest, `llmman rm m@sha256:...` deletes one of them in readdir order and prints `Removed docker.io/ai/m@sha256:...`, which is not the entry it removed. Refuse when a digest matches more than one entry and none is the spelled tag, listing them, the way the container tools do. At minimum print the stored ref that actually went.
2. The digest branch in `ref_matches_precise` compares `repo_name(stored) == repo_name(base)`, so it has no tag-only fallback like the branch below it. `find("m@sha256:...")` against a stored `docker.io/ai/m:v9` misses. Every caller resolves first today, so this is latent, but say so in the doc comment.

Also: this is four changes in one PR (store digest matching, the lock fold, the unload fallback for a lost tag, and the `daemon::unload` 404 envelope). The last one is independent and could land on its own.

ericcurtin:
Merged. Ran the suite, clippy and fmt locally, all clean.

Follow-ons, separate PRs:

1. `check_running_by_digest` and `unload_model`'s `min_by_key` both walk a `HashMap`, so with two running tags at one digest and neither spelled, which entry is picked is arbitrary. `remove` refuses that case. Make the other two deterministic or refuse the same way.
2. The PR body still describes the `daemon::unload` 404 envelope change, which is not in the diff. Send it on its own.
3. The `cp` digest-destination guard is unrelated to the bug and arrived without a mention in the title. Fine to keep, worth a line in the docs.
4. Still open from the description: a first load of `m:v9` racing a first load of its digest, since which tag holds a digest is unknown before the store is read.

PR #377, opened 2026-09-02
Forward explicit context length to vLLM

LxYuan0420:
This PR is:

- To pass explicit `LLMMAN_CONTEXT_LENGTH` values to vLLM as `--max-model-len`.
- To keep vLLM using its model-derived default when no context length is explicitly set.
- To document how the context setting maps to llama.cpp and vLLM.

When a user sets:

```text
LLMMAN_CONTEXT_LENGTH=4096
```

llmman already passes that value to llama.cpp as:

```text
--ctx-size 4096
```

This change passes the same explicit setting to vLLM as:

```text
--max-model-len 4096
```

When `LLMMAN_CONTEXT_LENGTH` is unset or invalid, llmman does not pass `--max-model-len` to vLLM, so vLLM keeps its normal model-derived default.


ericcurtin:
This looks like an easy merge to me, being I'm curious how:

```
llmman launch --model qwen3.8:safetensors claude
```

would work with vLLM. Most of the testing is done with llama-server from llama.cpp for speed reasons. But it would be interested to see if the long load times of vLLM would cause issues and what could be done there.

ericcurtin (commented):
Right idea and the split-out helpers are testable. Four things:

1. Rejecting `0` is an undocumented behaviour change for llama-server. `LLMMAN_CONTEXT_LENGTH=0` used to reach it as `--ctx-size 0`, meaning "use the model's trained context", and it also suppressed the OOM shrink retries. It now falls back to the VRAM-tiered default. Either keep `0` for llama-server and only filter it for vLLM, or say so in the docs.
2. `context_length_from_env` and `Inner::ctx_size` lost their doc comments. The ceiling behaviour and the pointer to `hostgpu::default_ctx_size` are still true and still worth reading. Put them back.
3. vLLM aborts when `--max-model-len` exceeds the model's `max_position_embeddings`, where llama-server caps down and warns. The table should say the two behave differently on an oversized value.
4. `vllm_max_model_len` deliberately skips `backend_ctx_size`'s `num_parallel` scaling, correctly, since vLLM's limit is per request. Say that in the doc comment so nobody "fixes" it.

As an aside, I'm curious how:

```
llmman launch --model qwen3.8:safetensors claude
```

would work with vLLM. Most of the testing is done with llama-server from llama.cpp for speed reasons. But it would be interested to see if the long load times of vLLM would cause issues and what could be done there.

LxYuan0420:
right, i missed a few details here. a couple of things:

- `LLMMAN_CONTEXT_LENGTH=0` should stay as `Some(0)` internally, so llama-server can still receive `--ctx-size 0`.
- vLLM should not receive `--max-model-len 0`; for vLLM, `0` means we omit the flag and let vLLM use its model/default behavior.
- i restored the env/docs wording so this is clearer from `llmman serve --help` and the README.
- i also added the note that llama-server and vLLM handle oversized context values differently: llama-server may cap, while vLLM rejects invalid values.

follow-up commits are pushed now.

docs for future self:

| `LLMMAN_CONTEXT_LENGTH` | llama-server gets | llama-server meaning | vLLM gets | vLLM meaning |
| --- | --- | --- | --- | --- |
| unset / invalid | no explicit user value | llmman/backend default behavior | no `--max-model-len` | vLLM uses model/default behavior |
| `0` | `--ctx-size 0` | llama-server uses model/default context | no `--max-model-len` | vLLM uses model/default behavior |
| positive value, e.g. `4096` | `--ctx-size 4096` before any GGUF parallel-slot scaling | requested context length for llama-server | `--max-model-len 4096` | requested max model length for vLLM |

ericcurtin:
Merged. Follow-ons, separate PRs:

1. Two curly apostrophes snuck into the `LLMMAN_CONTEXT_LENGTH` row in `docs/configuration.md`. Everything else there is ASCII.
2. That cell is now the longest in the table. Consider moving the per-backend mapping into a short table of its own.
3. `mlx_lm.server` still ignores the setting entirely. It has `--max-kv-size`, worth wiring up the same way.
4. Open question from the review: how `launch --model <ref>:safetensors` behaves against vLLM's load times, and whether the readiness wait needs a longer budget for that engine.

LxYuan0420:
@ericcurtin Partial answer for now: with qwen3.5:0.8b-safetensors, llmman launch did wait for vLLM to be ready before starting the integration.

My current machine is not a clean Claude Code test setup because of corporate account/env state, so I may need help testing the full Claude path on another machine.

ericcurtin:
@LxYuan0420 cool, thanks for the response

LxYuan0420:
@ericcurtin Thoughts after digging into the MLX side: `LLMMAN_CONTEXT_LENGTH` is token-based, so it maps cleanly to `llama-server --ctx-size` and `vLLM --max-model-len`.

For MLX, `--prompt-cache-bytes` is different: it limits retained prompt/KV cache memory in bytes. So I do not think we should map `LLMMAN_CONTEXT_LENGTH=4096` to `--prompt-cache-bytes 4096`; that would mean 4096 bytes, not 4096 tokens.

The clean follow-up would be a separate MLX knob like `LLMMAN_MLX_PROMPT_CACHE_BYTES`, but I think it is lower priority since it is more of an operator memory-pressure knob than something most users would tune.

Refs:
- https://github.com/ml-explore/mlx-lm/issues/615
- https://github.com/ml-explore/mlx-lm/pull/906

PR #383, opened 2026-09-03
fix(serve): route push 404 through AppError

maxgio92:
Follow-up to #338: the push handler's model-not-found 404 still hand-built its response body.
Routes it through `AppError::status`, same as the two 400s #338 already converted.
No status code or JSON body change.


ericcurtin:
Merged. Follow-ons:

- Assert the JSON body in the test, not just the status.
- `handle_pull`'s streaming error frames at serve.rs:4621 still hand-build `{"error"}`; worth a shared helper.

PR #387, opened 2026-09-03
test: rename has_host tests to describe resolve behavior

maxgio92:
Follow-up to #354: `has_host` was deleted, but its two former unit tests kept the old name.
Renames `has_host_requires_a_slash` to `resolve_requires_a_slash_to_treat_a_component_as_a_host`, and `has_host_recognizes_an_explicit_port` to `resolve_recognizes_an_explicit_port_as_a_host`.
Pure rename, no logic or assertion changes.

### DIFF
--- a/src/shortnames.rs
+++ b/src/shortnames.rs
@@ -936,7 +936,8 @@ mod tests {
         assert!(is_bare("gemma4:e4b"));
         assert!(!is_bare("unsloth/gemma4"));
         // Dots alone (no "/") no longer disqualify bareness — see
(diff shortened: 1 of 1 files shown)


ericcurtin:
Merged. Follow-ons:

- `shortnames.rs:196` still links `[`has_host`]`, a broken intra-doc link. Point it at `is_host_component`.
- Same stale name in comments at `shortnames.rs:846`, `:1019`, `:1030`.
- `verify.rs:333` keeps a local `has_host` with `first == "localhost"` while `is_host_component` is case-insensitive. Two spellings of one rule; fold them into one helper.

PR #396, opened 2026-09-03
feat(config): merge every config file into one llmman.conf

ericcurtin:
API keys could only be set through environment variables. This adds a config file for them, and — since that would have made three config files — folds the existing two into it.

## `llmman.conf`

```toml
# ~/.config/llmman/llmman.conf

[aliases]                        # was shortnames.conf
gemma4 = "docker.io/ai/gemma4"

[providers.openrouter]           # new
api_key = "sk-or-..."

[verify]                         # was verify.conf
default = "off"

[[verify.trust]]
pattern = "docker.io/myorg/**"
keys    = ["keys/myorg.pub"]
mode    = "enforce"
```

Read from `/etc/llmman/` then `~/.config/llmman/`, on every platform, with no environment variable to move them. `shortnames.conf` and `verify.conf` are no longer read, nor are the `/usr/share/llmman/`, `<binary>/../share/llmman/` and `<binary-dir>/` tiers — nothing ever shipped a file to them, and they are world-readable, which this file cannot be.

## Provider keys

Keyed by the provider id `llmman providers` prints, not by the environment variable — that name is models.dev's, not llmman's.

The environment still wins, as it does for `aws` and `gh`: the file is the standing answer, an `export` the deliberate this-session-only override. The reverse would make `OPENAI_API_KEY=... llmman run` silently spend the wrong key.

Resolution goes through one choke point (`providers::key_for`), so the CLI and daemon pick a key up identically, and every "no API key" message now names both places to put one. `llmman providers`' `set (shell)` became `set (client)`, since a client-side key no longer has to come from a shell.

## One file, three failure policies

A parse failure is not the same event for every reader. `verify` refuses to run — a trust policy llmman cannot read must not be mistaken for one that does not exist. Aliases and keys degrade to none, surfacing as "unknown reference" or "no API key". The error is printed once, centrally, so one typo is not announced three times.

## Permissions

A file carrying a key must not be readable by group or other, the way `ssh` refuses a loose private key. The check gates **the keys alone, not the read** — a world-readable `/etc/llmman/llmman.conf` still supplies its aliases and trust policy, which are not secrets.

## Testing

`cargo fmt --check`, `cargo clippy --release --all-targets -D warnings`, `cargo clippy --features fuzzing`, `cargo check` on the fuzz crate, and `cargo test --release --lib --bins` (536 passed) all clean.

Every pre-existing `verify` test passes unchanged — `Policy::parse` still takes a `[verify]` section body — so nothing silently weakened a security test. A new test drives a whole nested `llmman.conf` through to a `Policy`, so the two shapes cannot drift.

`config::load()` is `cfg!(test)`-guarded to return nothing, so no unit test can depend on whoever runs it having an `llmman.conf`; without that, a fixture provider named `openai` would resolve a real key and flip the assertions that prove no key leaks.

Smoke-tested against the release binary, since that path is `cfg!(test)`-guarded out:

- `[aliases]` resolves a short name from `~/.config/llmman/llmman.conf`
- malformed conf → one warning, `llmman verify` exits 1 (does **not** silently go `off`)
- daemon at `0600` → `llmman providers` reports the key `set`, with `OPENROUTER_API_KEY` unset
- same file at `0644` → `unset`, keys ignored with `chmod 600` advice, aliases unaffected
- `$XDG_CONFIG_HOME` and `~/Library/Application Support/llmman` are both ignored
- unknown section (`[alises]`) rejected rather than ignored

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ llmman push docker.io/myorg/mymodel:v1 --sign-key signing.key
 llmman verify docker.io/myorg/mymodel:v1 --key signing.pub
 ```
 
-A `verify.conf` trust policy turns that into an automatic check on every
+A `[verify]` trust policy turns that into an automatic check on every
 `pull`, warning or refusing outright per repository. Off by default —
 there is nothing to check against until you have said whom you trust.
 See [docs/verification.md](docs/verification.md).
@@ -261,19 +261,31 @@ broken command.
 
 All four commands ask the daemon over [`/llmman/providers`](#serve)
 rather than fetching models.dev themselves, so the cache outlives any
-single command and the key status reported is that of the environment
-whose key actually gets spent.
+single command and the key status reported is that of the process whose
+key actually gets spent.
 
 Requests still go through `llmman serve`; `--provider` changes where the
 daemon forwards them, not who the client talks to. So one endpoint and
 one place integrations are configured, whether a model is local or
 hosted, and both usable from the same session.
 
-The API key is read from the variable models.dev names for that provider
-and travels per request, never to disk. `hermes` is the exception: llmman
-configures it through a file on disk, so it can't carry a key and
-`llmman serve` needs the variable in its own environment instead. That
-fallback is only used for a daemon bound to loopback, and never for a
+The API key comes from the variable models.dev names for that provider,
+or — when that is unset — from `~/.config/llmman/llmman.conf`, keyed by
+provider id:
+
+```toml
+[providers.openrouter]
+api_key = "sk-or-..."
+```
+
+Either way it travels per request; llmman itself never writes a key
+anywhere. A file carrying one must be `chmod 600` or the keys in it are
+ignored with a warning, and an `export` still overrides it. See
+[docs/configuration.md](docs/configuration.md#provider-api-keys).
+
+`hermes` is the exception: llmman configures it through a file on disk,
+so it can't carry a key and `llmman serve` needs one of its own instead.
+That fallback is only used for a daemon bound to loopback, and never for a
 browser request from another site. It bounds the blast radius rather
 than authenticating anyone, so on a shared machine prefer an integration
 that sends its own key. `cline`, `kimi`, `copilot` and `openclaw` can't be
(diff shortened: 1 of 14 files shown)


ericcurtin:
Addressed the review. Summary of what I took and what I declined:

**Fixed**

- **Keys printable via `Debug`** — `ProviderConf` derived `Debug`, and `Conf`/`File` are public and derive it too, so `{:?}` on a `File` rendered plaintext keys. Replaced with a hand-written redacted `Debug`, matching `RemoteTarget` in `cmd::serve`. Test: `debug_output_never_carries_a_key`.

- **`wafer.ai` was unconfigurable** — confirmed against `tests/fixtures/models-dev-providers.json`; it is a real catalog id. `[providers.wafer.ai]` parses as two nested tables, so the hint told users to write something that silently configured nothing. `key_hint` now quotes ids that are not bare TOML keys (`[providers."wafer.ai"]`). Test: `key_hint_quotes_a_provider_id_that_is_not_a_bare_toml_key`.

- **A blank key could not shadow `/etc`** — `provider_keys()` dropped blanks per-file, so `api_key = ""` in the user file left the system key in force, contradicting documented later-wins precedence and leaving no way to opt out of a system-wide credential. Blanks now survive the merge and are dropped once at the end, so none is ever spent. Split `merge_keys` out to test without a filesystem. Test: `a_later_blank_shadows_an_earlier_key_and_is_never_spent`.

- **`/etc/llmman` was drive-relative on Windows** — a bare leading `/` resolves against the current drive, so launching from `D:` read `D:\etc\llmman`. Now anchored to `%SystemDrive%`. The search-path test compares against `system_dir()` and asserts it is absolute.

- **"regardless of the environment" was false** — `dirs::home_dir()` reads `$HOME` on Unix, which my own smoke test relied on. Corrected in the module docs and `docs/configuration.md` to "no llmman-specific variable (`~` is `$HOME` as usual)".

**Declined, with reasons**

- **Windows ACL validation** — correct that the owner-only default does not hold for a pre-existing widened ACL or for the system path. Reading a security descriptor needs a `windows` crate dependency this crate does not carry, which is more than this PR should take on. I replaced the overclaiming comment with an explicit statement of the limitation rather than leaving it implied.

- **`key_hint` showing a client path for a daemon-owned key** — real but narrow: it needs a healthy daemon running as another user. The message already scopes itself with "Where the daemon runs, ...", and plumbing a config path over `/llmman/providers` to fix the wording is not worth the wire-format change here.

All checks were green before these changes, including both Windows targets after the `cfg(unix)` dead-code fix that caused the earlier clippy failure. 539 unit tests pass.

Issue #324, opened 2026-08-29
serve --ociman: llama-server container gets no CPU limit


maxgio92:
Split out of #323 (raised in [review](https://github.com/llmmanorg/llmman/pull/323#pullrequestreview-5058942678)).

With `llmman serve --ociman`, [`container::spawn`](https://github.com/llmmanorg/llmman/blob/5c5b9366ffa9c6d143e5a49f6c1272a29e4be25e/src/container.rs#L232) builds the `docker run`/`podman run` command with no CPU limit. The llama-server container runs unconstrained even when the llmman daemon itself has a cgroup quota.

If a user constrains that container some other way (runtime defaults, pod limits), #322 recurs one level down. llama-server autodetects threads from all host cores and over-subscribes its quota. The `--threads` derivation from #323 does not apply there: the daemon's cgroup says nothing about the child container's.

Possible fix: forward the daemon's own quota as `--cpus` on the child container. Alternative: derive `--threads` from an explicitly configured limit. Either way `--ociman` needs a decision on its resource semantics.

PR #406, opened 2026-09-04
feat(serve): honor per-request options.num_thread

### LINKED ISSUE #328: serve: honor per-request options.num_thread
Split out of #323 (raised in [review](https://github.com/llmmanorg/llmman/pull/323#pullrequestreview-5060453208)).

llmman resolves `--threads` once at daemon startup. The value applies to every model. The Ollama API exposes `options.num_thread` per request; llmman ignores it.

Fix: pass it through the per-model plumbing that already carries `num_ctx`. An explicit request option wins over the derived default.

maxgio92:
Fixes #328.

A request's Ollama `options.num_thread` now wins over `LLAMA_ARG_THREADS`, the derived startup value, and llama-server autodetection. It is wired through [`ensure_model`](https://github.com/llmmanorg/llmman/blob/a55fbe01cbdb0ae2ad3705f3814034b72ff3a7c4/src/cmd/serve.rs#L3397) as a `request_threads` parameter, like `num_ctx`, and applied via [`request_threads.or(state.0.threads)`](https://github.com/llmmanorg/llmman/blob/a55fbe01cbdb0ae2ad3705f3814034b72ff3a7c4/src/cmd/serve.rs#L3604). The option is emitted as `--threads`, and llama.cpp applies env before CLI, so the flag wins over `LLAMA_ARG_THREADS`.

It applies on fresh loads only: a running instance is reused as-is, mirroring `num_ctx`.

The OpenAI-compatible, Anthropic, and transcriptions surfaces pass nothing: they have no Ollama options blob.

Under `--ociman` the value is deliberately ignored, like the derived one: neither describes the container's own CPU limits. The rationale lives on the [`LlamaOptions::threads` field doc](https://github.com/llmmanorg/llmman/blob/a55fbe01cbdb0ae2ad3705f3814034b72ff3a7c4/src/container.rs#L263); container CPU limits are tracked in #324.

A dedicated [`opt_num_thread`](https://github.com/llmmanorg/llmman/blob/a55fbe01cbdb0ae2ad3705f3814034b72ff3a7c4/src/cmd/serve.rs#L6983) extractor rejects zero, negatives, fractions, strings, and above-u32 values (the existing `opt_u32` cast would truncate), with [table-driven tests](https://github.com/llmmanorg/llmman/blob/a55fbe01cbdb0ae2ad3705f3814034b72ff3a7c4/src/cmd/serve.rs#L11513).

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -3373,10 +3373,28 @@ async fn resolve_remote_target(
 /// `headers` are the incoming request's, used only to pick up a caller-
 /// supplied provider API key (see [`resolve_remote_target`]); `None` from
 /// a surface that has none to offer.
+///
+/// `request_threads` is the caller's Ollama `options.num_thread` (see
+/// [`opt_num_thread`]); `None` from every surface without an Ollama
+/// options blob (OpenAI-compat, Anthropic, embeddings, preload).
+/// Precedence for the thread count a local llama-server ends up with:
+/// request option > `LLAMA_ARG_THREADS` > the derived `state.threads`
+/// (see [`threads_from_env_or_host`]) > llama-server's own
+/// autodetection. The request option is forwarded as `--threads`, and
+/// llama.cpp applies env before CLI, so that flag wins over
+/// `LLAMA_ARG_THREADS`; when the option is absent, `state.threads` is
+/// already `None` whenever `LLAMA_ARG_THREADS` is set, so no
+/// `--threads` is emitted and the env choice stands untouched. Like a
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):
LGTM. Follow-ons, not blockers:

1. A running instance is reused even when the request's num_thread differs from what it was started with. Compare the request's runner options against the loaded ones and reload on a mismatch, so a new num_thread takes effect without a manual unload.
2. Same check should cover num_ctx once per-request num_ctx lands.

PR #425, opened 2026-09-06
Treat only a model-not-found 404 as a missing model in llmman stop

ricky-chaoju:
`daemon::unload` read any 404 as "no such model", so a proxy in front of the daemon, or anything else answering on its port, turned into `couldn't find model "x" to stop`. It now takes only the `{"error":"model '…' not found"}` envelope the daemon sends as that answer and reports any other 404 with its status and body. A test renders `unload_model`'s own 404 through `into_response` and checks the client side accepts it, so the two ends cannot drift apart. Verified against a stand-in server answering 404 with an HTML body and with a foreign JSON body: both come back as `server returned 404 Not Found` with the body, and `llmman stop zz/nope` still reports the missing model.

follow up #376

### DIFF
--- a/src/cmd/serve.rs
+++ b/src/cmd/serve.rs
@@ -10747,7 +10747,10 @@ mod tests {
             .await
             .unwrap();
         assert!(
-            crate::daemon::is_model_not_found_body(std::str::from_utf8(&body).unwrap()),
+            crate::daemon::is_model_not_found_body(
+                std::str::from_utf8(&body).unwrap(),
+                "docker.io/ai/nothing-here"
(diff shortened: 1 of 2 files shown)


ericcurtin (approved):
Envelope-parsed 404 check matches how the reference `stop` does it, tests pin both ends. Merging.

Follow-ups:
- `llmman stop` for a hosted/provider model could short-circuit like the reference does for cloud models (no unload round-trip)
- the same "any 404 == not found" assumption may exist in other `daemon::` client helpers; worth a sweep

PR #428, opened 2026-09-06
Simplify the nvm lookup and trim the qwen doc comments

ricky-chaoju:
`find_qwen` now looks under `~/.nvm/versions/node/*/bin` with one directory scan, the way ollama does; the `NVM_DIR` and `XDG_CONFIG_HOME` variants and the version sort are gone. The doc comments on `launch_qwen`, `write_qwen_settings_at` and `qwen_settings_merged` are cut to a few lines each. The pre-flight e2e's elapsed bound is 30s, like the dead-daemon test, so it does not flake next to the model tests.

follow up #379


ericcurtin (approved):
Matches the reference nvm glob, CI green. Merging.

Follow-ups worth considering:
- the other launch integrations (claude/opencode/codex/hermes) still have their own PATH-probe lists; worth unifying the nvm lookup into one helper
- `qwen_settings_merged` comment still cites file paths in Qwen Code's tree that will drift; a link to the concept is enough

PR #430, opened 2026-09-06
docs: add Docker Compose deployment example

### LINKED ISSUE #364: docs: compose deployment example (serve + webui + store volume)
There is no compose file or deployment example in the repo. Someone who wants to
run `llmman serve` behind a gateway, with a webui alongside it and a persistent
store volume, has to work it out from scratch.

Scope: a documented compose example covering serve plus webui plus a named
volume for the model store, and a note on the CPU limit issue in #324.

@adq12, you built an API gateway and docker-compose stack for Ollama in
https://github.com/ollama/ollama/pull/16542. No obligation, but if you feel like
adapting the deployment shape here, or telling us what you would leave out, that
would be welcome.

be-student:
Closes #364

Adds a runnable Compose example with Caddy in front of `llmman serve`, the built-in web UI, and a named model-store volume. The image supports amd64 and arm64, verifies both downloaded binaries, and documents the `--ociman` CPU-limit caveat from #324.

Validation:
- `docker compose config`
- live `/api/version` and web UI requests through Caddy
- arm64 and amd64 image builds and `llama-server --version`
- writable persistent volume and four-CPU cgroup limit

AI assistance: I used Codex to implement and validate this change, then reviewed the final diff.

### DIFF
--- a/docs/compose.md
+++ b/docs/compose.md
@@ -1,0 +1,57 @@
+# Docker Compose
+
+[`examples/compose`](../examples/compose) runs `llmman serve` behind a Caddy
+gateway. The same gateway exposes the built-in web UI and the Ollama, OpenAI,
+and Anthropic-compatible APIs. A named volume keeps pulled models and the
+downloaded `llama-server` between container replacements.
+
+From the repository root:
+
+```sh
+docker compose -f examples/compose/compose.yaml up --build
+```
+
+Open <http://localhost:8080/> for the web UI. Clients can use the same address
+as their API base URL. For example:
+
+```sh
+curl http://localhost:8080/api/version
+```
+
+The image includes checksum-verified, pinned llmman and llama.cpp CPU binaries.
+Override the llmman version at build time when needed:
+
+```sh
+LLMMAN_VERSION=0.1.336 docker compose \
+  -f examples/compose/compose.yaml build --pull
+```
+
+To update llama.cpp, change `LLAMA_CPP_VERSION` and both architecture checksums
+in the Dockerfile together. A mismatched archive fails the image build.
+
+Bundling the pinned backend avoids depending on GitHub's rate-limited release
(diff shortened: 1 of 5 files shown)


ericcurtin (commented):
Thanks, this is close. Two small doc fixes and it's good to go. Converting to draft; please mark ready for re-review once updated.

FYI I checked the b10293 archive: it needs libssl3/libcrypto3 (pulled in via ca-certificates) but not libcurl, so the Copilot libcurl comment can be ignored.

docs/compose.md
```diff
@@ -0,0 +1,57 @@
+# Docker Compose
+
+[`examples/compose`](../examples/compose) runs `llmman serve` behind a Caddy
+gateway. The same gateway exposes the built-in web UI and the Ollama, OpenAI,
+and Anthropic-compatible APIs. A named volume keeps pulled models and the
+downloaded `llama-server` between container replacements.
```
ericcurtin: Not accurate: `llama-server` is baked into the image at `/opt/llama.cpp`, nothing is downloaded. Say the volume keeps pulled models.

docs/compose.md
```diff
@@ -0,0 +1,57 @@
...
+  -f examples/compose/compose.yaml build --pull
+```
+
+To update llama.cpp, change `LLAMA_CPP_VERSION` and both architecture checksums
+in the Dockerfile together. A mismatched archive fails the image build.
+
+Bundling the pinned backend avoids depending on GitHub's rate-limited release
+API during startup. The defaults match versions exercised by this repository.
+
+The `llmman-data` volume is mounted at `/var/lib/llmman`. Its model store is
```
ericcurtin: Wrong: the download cache lives under `$HOME/.local/share/llmman`, which is outside this volume. Drop the "backend binaries" claim.

PR #442, opened 2026-09-07
fix(store): canonicalize paths before digesting to prevent root escapes and duplicates

### LINKED ISSUE #362: store: canonicalize paths before digesting (symlinks, bind mounts, Windows mount points)
Nothing in `src/storage/` or `src/sources/mod.rs` calls `canonicalize`,
`symlink_metadata`, or `read_link`. Model files reached through a symlink, a
bind mount, or a Windows volume mount point are therefore digested and recorded
by their apparent path.

Consequences worth confirming with tests: the same blob imported by two paths
can be stored twice, and a path that escapes the store root is not detected.

Scope: resolve paths before digesting, and reject paths that resolve outside the
expected root.

@Saad-Mallebhari, you fixed exactly this class of bug in Ollama in
https://github.com/ollama/ollama/pull/17608, resolving NTFS volume mount points
and symlinks in `fileDigestMap`. No obligation, but if you want to port the idea
or just point out what we will get wrong on Windows, that would help.

Saad-Mallebhari:
Fixes #362

This PR addresses the Windows path canonicalization vulnerabilities when importing models through symlinks, bind mounts, or Windows directory junctions.

### The Problem
Previously, `WalkDir` in `sources/local.rs` and `OciStore::build` traversed symlinks and junction points but recorded the files by their **apparent paths**. This meant:
1. Duplicate blob storage if the same model was imported via different symlinks.
2. A malicious symlink inside the directory could quietly escape the store root and package arbitrary system files into the model image.

Additionally, using `std::fs::canonicalize` on Windows introduces verbatim path prefixes (`\\?\`), which break `strip_prefix` and other path logic downstream.

### The Fix
This mirrors the approach used in `ollama/ollama#17608`:
- Replaced `std::fs::canonicalize` with `dunce::canonicalize` globally (`fsutil.rs`, `cmd/build.rs`, `modelpack.rs`, `verify.rs`, `cmd/serve.rs`).
- Pre-canonicalize the source root before scanning.
- Canonicalize each traversed entry and strictly check `canonical_entry.starts_with(&canonical_root)` to prevent root escapes.
- Added comprehensive NTFS directory junction (`mklink /J`) tests in `sources/local.rs` to verify that root escapes fail and duplicate imports resolve correctly without requiring Developer Mode on Windows CI runners.

### DIFF
--- a/src/sources/local.rs
+++ b/src/sources/local.rs
@@ -15,19 +15,44 @@ pub(crate) fn pull(local_path: &str, target: &Target<'_>) -> Result<()> {
         anyhow::bail!("local path {local_path:?} is not a directory");
     }
 
+    // Canonicalize the root so that nested symlinks, junctions, and volume
+    // mount points can be validated against the real directory root.
+    // `dunce::canonicalize` resolves NTFS junctions and volume mount points
+    // while avoiding the `\\?\` verbatim-path prefix on Windows (which would
+    // break `strip_prefix` and `starts_with` comparisons downstream).
+    let canonical_root = dunce::canonicalize(root)
+        .with_context(|| format!("canonicalize local path {local_path:?}"))?;
+
     if target.report_cached(local_path, local_path) {
         return Ok(());
     }
 
     let mut packed = Vec::new();
-    for entry in walkdir::WalkDir::new(root).follow_links(false) {
+    for entry in walkdir::WalkDir::new(&canonical_root).follow_links(true) {
         let entry = entry.with_context(|| format!("walk {local_path}"))?;
+
+        // Canonicalize each entry to resolve any nested symlinks/junctions and
+        // verify it hasn't escaped the source root. A symlink or directory
+        // junction pointing outside canonical_root is rejected immediately.
+        let canonical_entry = dunce::canonicalize(entry.path()).with_context(|| {
+            format!("canonicalize {} under {local_path}", entry.path().display())
+        })?;
+        if !canonical_entry.starts_with(&canonical_root) {
+            anyhow::bail!(
+                "path {:?} resolves to {:?} which is outside store root {:?}",
+                entry.path(),
+                canonical_entry,
+                canonical_root
+            );
+        }
+
         if !entry.file_type().is_file() {
             continue;
         }
+
         let rel = entry
             .path()
-            .strip_prefix(root)
+            .strip_prefix(&canonical_root)
             .unwrap_or(entry.path())
             .to_string_lossy()
             // Recorded in the manifest, read back on any platform:
@@ -37,7 +62,7 @@ pub(crate) fn pull(local_path: &str, target: &Target<'_>) -> Result<()> {
             continue;
         }
         packed.push(PackFile {
-            local_path: entry.path().to_path_buf(),
+            local_path: canonical_entry,
             relative_path: rel,
             owned: false,
         });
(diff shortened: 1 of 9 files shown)


src/sources/local.rs
```diff
@@ -55,13 +85,13 @@ pub(crate) fn pull(local_path: &str, target: &Target<'_>) -> Result<()> {
...
 
-    let repo = root
+    let repo = canonical_root
         .file_name()
         .map(|n| n.to_string_lossy().into_owned())
         .unwrap_or_else(|| local_path.to_string());
     super::pack_as_model_pack(
         target,
-        local_path,
+        &canonical_path_str,
```
ericcurtin, line 94: Recording the manifest under the canonical path breaks lookups: `cmd/resolve.rs:110` and serve still look the model up by the reference the user typed, so `llmman run /symlink/model` pulls, then fails to find it. Keep recording under `local_path` (blobs are content-addressed, so dedupe already happens), or canonicalize at those call sites too.

src/sources/local.rs
```diff
@@ -15,19 +15,49 @@ pub(crate) fn pull(local_path: &str, target: &Target<'_>) -> Result<()> {
...
+    // digests, instead of being stored twice.  `dunce::canonicalize`
+    // resolves NTFS junctions and volume mount points while avoiding the
+    // `\\?\` verbatim-path prefix that `std::fs::canonicalize` produces
+    // on Windows (which would break `strip_prefix` and `starts_with`
+    // comparisons downstream).
+    let canonical_root = dunce::canonicalize(root)
+        .with_context(|| format!("canonicalize local path {local_path:?}"))?;
+    let canonical_path_str = canonical_root.to_string_lossy().replace('\\', "/");
+
+    if target.report_cached(&canonical_path_str, local_path) {
```
ericcurtin, line 29: Same as below: the cache check should agree with whatever ref is recorded, or every run re-imports.

src/sources/local.rs
```diff
@@ -167,4 +197,104 @@ mod tests {
...
+                store_as: None,
+            },
+        )
+        .unwrap();
+
+        // Check the generated manifest. The canonical path (real_src) should be used as the repo name
+        // (unless overridden by store_as), avoiding duplicate blobs if pulled via multiple symlinks.
+        let canonical_str = dunce::canonicalize(&real_src).unwrap().to_string_lossy().replace('\\', "/");
+        let expected_repo = real_src.file_name().unwrap().to_string_lossy().into_owned();
+        let desc = oci::read_manifest_ref(&layout, &canonical_str).unwrap();
```
ericcurtin, line 254: Test asserts nothing; `expected_repo` and `desc` are unused (clippy `-D warnings` will fail). Assert the repo name / that a second pull via the link reports cached.

src/sources/local.rs
```diff
@@ -15,19 +15,49 @@ pub(crate) fn pull(local_path: &str, target: &Target<'_>) -> Result<()> {
...
+        .with_context(|| format!("canonicalize local path {local_path:?}"))?;
+    let canonical_path_str = canonical_root.to_string_lossy().replace('\\', "/");
+
+    if target.report_cached(&canonical_path_str, local_path) {
         return Ok(());
     }
 
     let mut packed = Vec::new();
-    for entry in walkdir::WalkDir::new(root).follow_links(false) {
+    for entry in walkdir::WalkDir::new(&canonical_root).follow_links(true) {
```
ericcurtin, line 34: `follow_links(false)` -> `true` is a behavior change worth a line in the commit body: file symlinks inside the root that point outside were skipped, now they error.

src/sources/local.rs
```diff
@@ -15,19 +15,49 @@ pub(crate) fn pull(local_path: &str, target: &Target<'_>) -> Result<()> {
...
         if !entry.file_type().is_file() {
             continue;
         }
-        let rel = entry
-            .path()
-            .strip_prefix(root)
+
+        // Canonicalize each entry to resolve any nested symlinks and
+        // verify it hasn't escaped the source root.  A symlink inside
+        // the model directory pointing to `../../etc/passwd` would
```
ericcurtin, line 42: CI: `cargo fmt --all` fails here and in the tests.

ericcurtin (changes requested):
Thanks, the escape check and `dunce` swap look right. A few things to address before merge; I'm moving this to draft, please mark ready for review again once CI is green.

ericcurtin (approved):
All prior points addressed: manifest and cache check stay keyed on the requested `local_path`, tests assert, `follow_links` change is called out in the commit body, fmt/clippy clean. Green on all five targets including both Windows legs. LGTM.

PR #449, opened 2026-09-08
Re-extract incomplete safetensors cache files

LxYuan0420:
This PR is:

- To treat a safetensors cache file as reusable only when its size matches the OCI layer size.
- To retry extraction when a cached file is missing or incomplete.
- To avoid keeping a broken cache file around until someone deletes the cache manually.

Before this change, `extract_safetensors_dir` skipped a cached destination as soon as the path existed. If a copy stopped partway through, the short file could stay in the cache and later loads would fail somewhere inside vLLM or MLX.

This change keeps the same fast path for completed files, but recopies when the cached file size does not match the layer it came from.

### DIFF
--- a/src/modelpack.rs
+++ b/src/modelpack.rs
@@ -586,7 +586,12 @@ fn extract_safetensors_dir(
             continue;
         }
         let dest = cache_dir.join(rel_path);
-        if dest.exists() {
+        // exists() is not complete: a killed copy can leave a short dest.
+        if dest
+            .metadata()
+            .map(|m| m.is_file() && m.len() == layer.size)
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):
LGTM, thanks. Follow-ons worth considering:

- Copy to a temp file next to `dest` and rename on success, so a killed copy never leaves a partial at the final path (same pattern as blob downloads).
- Avoid the copy entirely: hardlink (fallback symlink) `dest` to the blob. Weights are multi-GB and the store already holds the bytes; the cache should point at them, not duplicate them.

PR #460, opened 2026-09-09
Link safetensors cache files to OCI blobs

LxYuan0420:
This PR is:

- To follow up on #449.
- To avoid copying safetensors weights into the cache.
- To link cache files to the existing OCI blob instead.

#449 made llmman stop trusting incomplete cache files.

This follow-up keeps that check, but changes the happy path from copying the weight file to linking it. That avoids storing the same multi-GB model weights twice while keeping the cache layout that vLLM and MLX expect.


ericcurtin (approved):
Verified locally (clippy/fmt/tests on top of main). Hardlink + atomic-copy fallback and the nlink-aware GC accounting look right.

PR #465, opened 2026-09-09
Move the Ollama API handlers into serve/ollama.rs

ricky-chaoju:
Moves everything under Ollama's `/api/*` out of `serve.rs` into `serve/ollama.rs`: version, tags, ps, show, pull, push, copy, delete, the blob and create routes, chat, generate, embed and embeddings, together with the helpers that sit under those banners (the push progress relay, blob staging, the embedding batching). No tests move; the handler tests in `serve/tests.rs` build the full state and stay there, importing the eleven helpers they use. The 15 handlers the router mounts and those helpers are `pub(super)`; apart from one signature rustfmt re-wrapped, the moved lines are unchanged. The second commit updates the four doc comments that name the old path.


ericcurtin (approved):
Verified locally: pure move (only visibility + imports differ), clippy/fmt clean, serve tests pass.

PR #466, opened 2026-09-09
Add pool integration

leandroBorgesFerreira:
# Add pool integration to launch command

## Summary

Adds support for launching the [Poolside CLI](https://poolside.com) with `llmman launch pool`,
enabling developers to use pool with local or hosted models through llmman's unified interface.

### Implementation
- Added pool to the list of supported integrations in `src/cmd/launch.rs`
- Implemented `launch_pool()` function that:
  - Finds the `pool` binary on PATH
  - Sets the `POOLSIDE_STANDALONE_BASE_URL` environment variable to point to llmman's `/v1`
endpoint
  - Sets `POOLSIDE_API_KEY` to `"llmman"` for authentication
  - Passes the model via the `-m` flag
  - Forwards any extra arguments to pool

### Testing
- Added `launch_pool_with_model()` end-to-end test in `tests/launch_e2e.rs`
- Test verifies pool can successfully connect to llmman and respond to prompts
- Follows the same pattern as existing integration tests (claude, opencode, codex, etc.)

## Usage

```bash
# Launch pool with a specific model tag
llmman launch pool --model llama3.2:latest

# Launch pool with a hosted provider
llmman launch pool --provider anthropic --model claude-sonnet-4

# Pass extra arguments to pool
llmman launch pool --model qwen3.8:latest -- --verbose
```


src/cmd/launch.rs
```diff
@@ -1405,6 +1411,30 @@ fn qwen_entry_is_ours(entry: &serde_json::Value, base_url: &str) -> bool {
...
+        args.extend(["-m".to_string(), model.to_string()]);
+    }
+    args.extend_from_slice(extra_args);
+
+    exec_with_env(
+        &bin,
+        &args,
+        &[
+            ("POOLSIDE_STANDALONE_BASE_URL", base_url.as_str()),
+            ("POOLSIDE_API_KEY", "llmman"),
```
ericcurtin: Pass `api_key` through here (as `launch_qwen` does) instead of the literal, otherwise `--provider ...` from the PR description sends the placeholder key to the daemon.

src/cmd/launch.rs
```diff
@@ -1405,6 +1411,30 @@ fn qwen_entry_is_ours(entry: &serde_json::Value, base_url: &str) -> bool {
             .is_some_and(|u| u.trim_end_matches('/') == base_url.trim_end_matches('/'))
 }
 
+/// pool: Poolside CLI, pointed at our /v1 endpoint.
+///
+/// Sets POOLSIDE_STANDALONE_BASE_URL and POOLSIDE_API_KEY.
+fn launch_pool(model: &str, extra_args: &[String]) -> anyhow::Result<()> {
+    let bin = find_on_path("pool").ok_or_else(|| anyhow::anyhow!("pool is not installed"))?;
```
ericcurtin: `pool` has no Windows build; `bail!` on `cfg!(windows)` before the PATH lookup with a clear "not supported on Windows" message.

ericcurtin (changes requested):
Thanks, close to what's needed. A few small things below; also please add a `pool` row to the table in `docs/providers.md`. Setting to draft, mark ready when addressed and I'll re-review. CI was waiting on approval, it's running now.

.github/workflows/ci.yml
```diff
@@ -808,6 +808,12 @@ jobs:
         run: |
           & ([scriptblock]::Create((irm https://hermes-agent.nousresearch.com/install.ps1))) -SkipSetup
 
+      # hermes has no npm package (install.sh/install.ps1 only, like
+      # llmman's own install scripts). 
+      - name: Install pool
+        if: runner.os != 'Windows'        
+        run: POOL_INSTALL_ACCEPT_EULA=1 curl -fsSL https://downloads.poolside.ai/pool/install.sh | sh
```
leandroBorgesFerreira, line 848: @ericcurtin I would like your input here. Tracking the  version of the installer of pool would be super annoying and I don't think there's a versioned version of the installer. 

Maybe commit the installer?

.github/workflows/ci.yml
```diff
@@ -808,6 +808,12 @@ jobs:
         run: |
           & ([scriptblock]::Create((irm https://hermes-agent.nousresearch.com/install.ps1))) -SkipSetup
 
+      # hermes has no npm package (install.sh/install.ps1 only, like
```
ericcurtin: Comment is about hermes (copy-paste). Say pool has no npm package instead. Trailing whitespace on this line and the `if:` below.

tests/launch_e2e.rs
```diff
@@ -1054,6 +1054,49 @@ fn openclaw_pull_registry_flake(stderr: &str) -> bool {
...
+    if !on_path("pool") {
+        eprintln!("skipping: pool not on PATH — https://poolside.com");
+        return;
+    }
+
+    // `exec -p <prompt>`: pool's non-interactive one-shot mode.
+    // First verify pool responds with "pong" (basic integration test).
+    launch_and_assert("pool", &["exec", "-p", PROMPT]);
+
+    // Then confirm the model was actually used in the /v1 request by
```
ericcurtin: This can't fail: `warm_model()` already loaded `MODEL` via `llmman run` before the launch, so `ps` shows it regardless of what pool did. Drop the block; `launch_and_assert` is the test.

src/cmd/launch.rs
```diff
@@ -1405,6 +1411,58 @@ fn qwen_entry_is_ours(entry: &serde_json::Value, base_url: &str) -> bool {
...
+///
+/// Sets POOLSIDE_STANDALONE_BASE_URL and POOLSIDE_API_KEY.
+/// For `pool exec` (standalone mode), sets POOLSIDE_STANDALONE_MODEL.
+/// For interactive mode, passes -m flag.
+fn launch_pool(model: &str, api_key: &str, extra_args: &[String]) -> anyhow::Result<()> {
+    if cfg!(windows) {
+        anyhow::bail!("pool is not supported on Windows");
+    }
+
+    // When --provider is supplied (api_key is a real provider key, not the
```
ericcurtin: Nit: no other launcher guards against a user redirecting its own CLI; consider dropping this to keep `launch_pool` minimal (env + `-m` + args).

ericcurtin (changes requested):
Verified the launch for real: `llmman launch pool --model qwen3.5:0.8b -- exec -p "...pong"` against pool 1.0.16 + llama-server prints `pong`, exit 0. Confirmed `pool exec -m` is rejected as obsolete, so the `POOLSIDE_STANDALONE_MODEL` split is right. Clippy clean, CI green.

Still missing from the last review: the `pool` row in the `## Integrations` table in `docs/providers.md`. Plus the small items inline. Setting to draft; mark ready when addressed and I'll re-review.

Re the installer question: unpinned is fine, same as hermes.

.github/workflows/ci.yml
leandroBorgesFerreira: Comment fixed. Trailing whitespace removed.

src/cmd/launch.rs
leandroBorgesFerreira: Makes sense. Let's not overcomplicate this. Removed.

tests/launch_e2e.rs
leandroBorgesFerreira: Block dropped.

ericcurtin (changes requested):
Code is good to go: matches the expected launcher shape (Windows bail, PATH lookup, `POOLSIDE_STANDALONE_BASE_URL`/`POOLSIDE_API_KEY`, `-m`), plus the verified `exec` split. `docs/providers.md` row is in. Clippy, fmt and tests clean locally.

One thing: #400 just landed and adds its own install step at the same spot in `ci.yml`, so this now conflicts. Rebase and keep both steps (order doesn't matter). `launch.rs`, `docs/providers.md` and `launch_e2e.rs` auto-merge.

Moving to draft; mark ready once rebased and I'll merge.

.github/workflows/ci.yml
ericcurtin: No don't commit the installer

src/cmd/launch.rs
```diff
@@ -605,13 +610,14 @@ fn launch(
...
         "goose" => launch_goose(model, api_key, extra_args),
+        "pool" => launch_pool(model, api_key, extra_args),
         other => anyhow::bail!(
             "unknown integration {:?}\nRun 'llmman launch' without arguments to list supported integrations.",
             other
         ),
     }
 }
-
+
```
ericcurtin, line 620: Trailing whitespace here is what fails `cargo fmt --check` on x86_64 Linux. Make it an empty line.

.github/workflows/ci.yml
```diff
@@ -972,6 +972,11 @@ jobs:
...
           Rename-Item (Join-Path $installDir 'antigravity.exe') 'agy.exe'
           $installDir | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      # pool has no npm package (install.sh/install.ps1 only, like
+      # llmman's own install scripts).
+      - name: Install pool
+        if: runner.os != 'Windows'
+        run: POOL_INSTALL_ACCEPT_EULA=1 curl -fsSL https://downloads.poolside.ai/pool/install.sh | sh
 
       # opencode installs its configured provider's npm package (here
       # @ai-sdk/openai-compatible — see launch::opencode_config) on demand
```
ericcurtin, line 982: Installs to `~/.local/bin`, which only reaches PATH via the goose step (and only if that succeeds). Add `echo "$HOME/.local/bin" >> "$GITHUB_PATH"` or the test silently skips.

ericcurtin, line 982: Pin it: the installer takes a version, `| sh -s -- v1.0.16`. Every other CLI in this job is pinned.

.github/workflows/ci.yml
```diff
@@ -972,6 +972,11 @@ jobs:
           Expand-Archive $archivePath -DestinationPath $installDir -Force
           Rename-Item (Join-Path $installDir 'antigravity.exe') 'agy.exe'
           $installDir | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+      # pool has no npm package (install.sh/install.ps1 only, like
+      # llmman's own install scripts).
+      - name: Install pool
+        if: runner.os != 'Windows'
+        run: POOL_INSTALL_ACCEPT_EULA=1 curl -fsSL https://downloads.poolside.ai/pool/install.sh | sh
```
ericcurtin, line 980: Nit: add `pool` to the e2e job name and the `claude|agy|...|goose` comment above.

tests/launch_e2e.rs
```diff
@@ -1198,6 +1198,25 @@ fn openclaw_pull_registry_flake(stderr: &str) -> bool {
...
+        eprintln!("skipping: llama-server not on PATH (required to serve any model)");
+        return;
+    }
+    if !on_path("pool") {
+        eprintln!("skipping: pool not on PATH — https://poolside.com");
+        return;
+    }
+
+    // `exec -p <prompt>`: pool's non-interactive one-shot mode.
+    // First verify pool responds with "pong" (basic integration test).
```
ericcurtin, line 1216: Stale comment from the dropped `ps` block; remove.

ericcurtin (changes requested):
Launcher itself looks right (env + `-m` + args, Windows bail; the `exec` branch is needed, `pool -m x exec` rejects `-m`). Remaining items are CI: the fmt failure is what's red, and the install step needs PATH + a pin. Setting to draft; flip back to ready when fixed and I'll re-review.

PR #468, opened 2026-09-09
feat: add DeepSeek Harness integration

### LINKED ISSUE #436: Add "llmman launch dsh"
And test aggressively to ensure it works before we open the PR.

HassanBahati:
## Summary

- Added dsh (DeepSeek Harness) to the llmman launch integration registry and dispatch path.
- Added focused unit tests for the generated settings/patch YAML and the --patch conflict guard.

Fixes #436


src/cmd/launch.rs
```diff
@@ -1857,6 +1987,123 @@ model = \"gpt-5\"
...
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap()
+                .as_nanos()
+        ));
+        let settings_path = dir.join("settings.yaml");
+        let patch_path = dir.join("llmman.cordis.yml");
+        write_dsh_patch(&patch_path, &settings_path).unwrap();
+        let contents = std::fs::read_to_string(&patch_path).unwrap();
+        assert!(contents.contains("id: settings"));
+        assert!(contents.contains(&format!("path: \"{}\"", settings_path.to_string_lossy())));
```
ericcurtin: Fails on Windows: `yaml_quote` escapes `\` so the raw path never matches. Compare against `yaml_quote(&settings_path.to_string_lossy())` instead.

src/cmd/launch.rs
```diff
@@ -1293,11 +1302,132 @@ fn qwen_entry_is_ours(entry: &serde_json::Value, base_url: &str) -> bool {
...
+        .join("llmman")
+        .join("launch")
+        .join("dsh"))
+}
+
+/// A directory unique to this launch. dsh's settings-file provider
+/// watches and hot-reloads its document, so a second concurrent
+/// `llmman launch dsh` writing one shared path would silently retarget
+/// an already-running instance's model — see `launch_dsh`'s cleanup.
+fn dsh_launch_dir() -> anyhow::Result<PathBuf> {
```
ericcurtin: Simpler to write fixed `~/.config/llmman/launch/dsh/settings.yaml` + `llmman.cordis.yml` on every launch (no per-launch dir, no cleanup hook in `exec_with_env`). Hot-reload retargeting only bites if two concurrent launches pick different models; acceptable trade-off for less code.

ericcurtin (changes requested):
Thanks for this. Windows CI is red; setting to draft, please mark ready for re-review once it's green.

tests/launch_e2e.rs
```diff
@@ -894,6 +894,30 @@ fn qwen_loop_detection(stderr: &str) -> bool {
...
+    }
+
+    // `--profile headless <prompt>`: dsh's own one-shot mode — answers
+    // one task, prints the final assistant message, and exits. The
+    // default `web` profile `llmman launch dsh` execs otherwise boots a
+    // persistent server with no text reply, so it alone wouldn't fit
+    // this file's launch_and_assert/"pong" pattern at all; a caller-
+    // supplied `--profile` overrides it (see `dsh_args` in
+    // `cmd::launch`) for exactly this case.
+    launch_and_assert("dsh", &["--profile", "headless", PROMPT]);
```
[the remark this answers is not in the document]
HassanBahati: True, I tried to match what's already shipped here: `write_codex_config` (`~/.codex`), `hermes_home`, and openclaw's onboarding check all use `dirs::home_dir()`, with no override set in `run_launch` either. qwen is isolated only because Qwen Code provides `QWEN_HOME` upstream. llmman seems to honour integrations' own vars rather than inventing its own.

Worth noting the two effects flagged here are ones the fixed-path design already accepts: that file is rewritten on every launch, and cross-launch retargeting was the stated trade-off.

I'm happy to add an override if you'd prefer  though I had kept it uniform to codex/hermes/openclaw/dsh than make dsh the only one carrying it.

cc @ericcurtin

ericcurtin (approved):
Verified against real dsh 0.1.2-rc.1: `web --patch` and `--profile headless` both pick up the settings overlay, `web` refuses a parent `--patch` as described. Ran `llmman launch dsh --model qwen3.5:0.8b -- --profile headless "...pong"` end to end locally against llama-server: prints `pong`, exit 0. Clippy clean, CI green on all targets.

PR #472, opened 2026-09-10
Move the OpenAI pass-through handlers into serve/openai.rs

ricky-chaoju:
Moves the "OpenAI pass-through handlers" block of `serve.rs` into `serve/openai.rs`: models, chat/completions, completions and embeddings, with the request preparation and relay helpers the media, transcription, Responses and Anthropic routes share with them (`proxy_openai_generation`, `proxy_openai_passthrough`, `forward_openai_request`). Those other `/v1` routes stay for now. No tests move; `serve/tests.rs` imports the two helpers it uses. The items `serve.rs` and the tests still name are `pub(super)`; the moved lines are otherwise unchanged.


ericcurtin (approved):
Verified pure move (line-by-line diff of removed vs added is only imports/module doc). Clippy -D warnings and `cargo test cmd::serve` pass on the merged tree.

PR #475, opened 2026-09-11
feat: add llmman launch goose

### LINKED ISSUE #335: Add "llmman launch goose"

HassanBahati:
## Summary
image

Adds `goose` (Block's agent) to `llmman launch`

goose reads `GOOSE_PROVIDER`/`GOOSE_MODEL` and the OpenAI provider's `OPENAI_HOST`/`OPENAI_BASE_PATH`/`OPENAI_API_KEY` in preference to its own
`config.yaml`, so `launch_goose` sets those five and writes nothing. Unlike
`hermes` and `qwen` there is no config to merge and no credential on disk, so
`--provider` works without the daemon holding the key.

## Notes

- **`MODEL_REQUIRED`**: goose is added for a reason of its own, so that const's
  doc comment now carries both. Without a model goose refuses with
  `Run 'goose configure' first` — advice that doesn't apply to a launch llmman
  configures through the environment — so llmman refuses first, before
  `ensure_server`.
- **`find_goose`** falls back to `~/.local/bin`, where `download_cli.sh`
  installs without putting it on `PATH`.
- **CI**: goose has no npm package, so `install_cli` can't reach it. A separate
  step installs from a pinned release (v1.50.0), following the hermes
  precedent. It warns rather than fails because goose publishes no
  `aarch64-pc-windows` asset — the Windows arm64 leg skips the test on its own
  `on_path` check.
- **`GOOSE_MODE`** is deliberately not set: auto-approving an agent's writes is
  the user's call, not llmman's. The e2e test sets it for itself.

## Testing

Against real goose 1.50.0 and a real served model:

- `llmman launch goose --model qwen3.5:0.8b -- run -t "..." --no-session`
  answers `pong`, with goose's banner confirming
  `openai docker.io/ai/qwen3.5:0.8b`
- the env-only design proven from a genuinely clean slate (no `~/.config/goose`
  at all), and `~/.config/goose/` still holds zero files afterwards
- `find_goose`'s fallback exercised with `~/.local/bin` stripped from `PATH`
- a missing `--model` refused before any daemon start
- `tests/launch_e2e.rs::launch_goose_with_model` runs the real CLI end to end
  (confirmed running rather than skipping)
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and the unit
  suite all clean

Fixes #335


ericcurtin (changes requested):
Implementation looks good: env-only configuration, no files written, `--provider` works, Windows install dir probed.

Needs a rebase: `ci.yml` conflicts with the vllm/sglang e2e additions on main (job name, `if:` with `workflow_dispatch`, and the timeout block, now 400 there). Keep main's version of those and re-add goose to the name and the test-count comment. `docs/providers.md` and `tests/launch_e2e.rs` auto-merge.

Moving to draft; mark ready once rebased and I'll re-review.

ericcurtin (dismissed):
Rebased cleanly, Windows install dir probed, pin moved to `GOOSE_VERSION`. fmt/clippy/tests green locally. LGTM.

ericcurtin (approved):
LGTM. Env-only configuration matches how the other agent launchers hand the endpoint and key to the CLI; no config written, no key persisted. fmt/clippy/unit suite green locally on top of current main.

PR #476, opened 2026-09-11
Move the media and transcription routes into serve/openai.rs

ricky-chaoju:
Moves the three `/v1` sub-sections that were still in `serve.rs` after #472 into `serve/openai.rs`: the media generation routes (`/v1/images/generations`, `/v1/videos`, `/v1/audio/speech`), the vLLM-Omni dialect behind them, and `/v1/audio/transcriptions`. They sit under their existing banners at the end of the module, and the module doc now names them. No tests move; `serve/tests.rs` imports the four helpers it uses. The six items the router mounts and those helpers are `pub(super)`; apart from one signature rustfmt re-wrapped and the `handle_openai_media` doc link no longer needing `super::`, the moved lines are unchanged. The second commit turns the two doc links in `serve.rs` that now point into the module into plain code spans.


ericcurtin (approved):
Pure move verified with `git diff --color-moved`; only visibility and import changes. LGTM.

PR #492, opened 2026-09-13
fix(storage): replace unwrap on strip_prefix with contextual error in OciStore::build

Saad-Mallebhari:
Follow-up to #442.

## Summary

This addresses the `unwrap()` concern raised during the review of #442 by replacing the `strip_prefix()` unwrap in `OciStore::build` (`src/storage/oci.rs`) with contextual error handling (`anyhow::Context`).

This ensures that an unexpected path discrepancy returns a descriptive error rather than panicking.

## Validation

- `cargo test --lib storage::oci::tests`
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --all -- --check`

All checks pass locally with zero warnings.

---

This PR is intentionally kept as a minimal, standalone follow-up to the original PR.

For the other follow-up mentioned in the review regarding external links, I'll keep that as a separate discussion/PR once the preferred direction is confirmed.

@ericcurtin For the other follow-up you mentioned around external links, would you prefer the `--allow-external-links`-style escape hatch, or a different approach for legitimate layouts such as the Hugging Face cache?

I can work on that as the second PR once I know which direction you'd prefer.


ericcurtin (approved):
LGTM.

PR #493, opened 2026-09-13
Move the Anthropic Messages routes into serve/anthropic.rs

ricky-chaoju:
Moves the "Anthropic /v1/messages" block still in `serve.rs` after #476 into the existing `serve/anthropic.rs`: the route handler, its fork between translating through `messages` and relaying to a provider that speaks the dialect, and the headers that relay passes through. It keeps its banner and sits ahead of the inline tests. The module doc now names the route and scopes its purity claim to the translation, since the handler resolves a target before it converts anything and the old sentence would otherwise have been false.

No tests move; `serve/tests.rs` imports `relay_anthropic_messages` from its new home and picks up `apply_default_repeat_penalty`, whose only caller in `serve.rs` was the moved code. The handler and the relay are `pub(super)`, `anthropic_messages_to` and `ANTHROPIC_PASSTHROUGH_HEADERS` stay private, and nothing in `serve.rs` or a sibling had to widen. Apart from `MESSAGES_ROUTE` losing its module prefix now that it is local, and the `Wire::Anthropic` doc link needing its full path in a module that does not import `Wire`, the moved lines are unchanged. The second commit points the `launch_e2e` doc comment at the new path.


ericcurtin (approved):
LGTM — pure move, CI green, no conflicts with main.

PR #498, opened 2026-09-14
Move the OpenAI Responses routes into serve/responses.rs

ricky-chaoju:
Moves the OpenAI Responses block left in `serve.rs` after #493 into the existing `serve/responses.rs`: both route handlers, the native-first fallback to a translated chat completion, and the two request-shape workarounds `llama-server` needs. The banner travels with it, and the module doc narrows its purity claim to the conversion now that a handler in the file goes to the network.

The cut splits a comment: #439 inserted `trait SseConverter` between `remote_responses`' doc and its function, so those ten lines travel back to the function and the trait keeps the two that were always its own. `SseConverter`, `sse_converter!` and `convert_upstream` stay, since `anthropic.rs` calls the last of them, as do `unsupported_on_wire`, `wire_refusal` and `explain_missing_route`, which `ollama.rs` and `openai.rs` call for other routes; both orphaned groups get a one-line banner in place of the heading that left. Apart from nine `pub(super)` widenings, three now-local paths and two doc links, the moved lines are unchanged.


ericcurtin (approved):
Approved.

Issue #429, opened 2026-09-06
Split src/cmd/serve.rs (13.5k lines)


ericcurtin:
`src/cmd/serve.rs` is 13,500 lines (~8,250 of code, ~5,250 of tests in `mod tests` from line 8256). It's the largest file in the repo by a wide margin and the one most people need to touch, so it's the biggest barrier to contributing. Split it.

lesbass:
Retracting my earlier claim of the `serve/config.rs` slice — it was already covered by #431 (merged 2026-09-07), so I've closed #467. Sorry for the noise.

PR #508, opened 2026-09-15
Move the Ollama option extractors into serve/ollama.rs

ricky-chaoju:
Moves the three Ollama options-blob extractors out of `serve.rs` into `serve/ollama.rs`. `ollama.rs` was their only caller, so nothing imports them back and all three become `pub(super)`; `serve/tests.rs` picks them up from the new home.

The three line section header becomes the one line `// -- Ollama options blob ---` form the destination file already uses seven times. `opt_num_thread`'s doc link from `ensure_model` is requalified, and the second commit retargets a prose citation in `cmd::run`. Otherwise the moved lines are unchanged: nineteen of the twenty-two are byte-identical, the other three being the visibility keywords.


ricky-chaoju:
@codex review

ericcurtin (approved):
LGTM. Pure move, CI green.

ericcurtin:
Merged, thanks. Follow-ups for #429 if you want to keep going:
- Move `options_to_oai` and the rest of the Ollama-options → llama-server mapping next to these extractors.
- `opt_u32` still truncates with `as u32`; consider `u32::try_from` like `opt_num_thread`.
- Add unit tests for `opt_f64`/`opt_u32` on non-numeric JSON values.

PR #510, opened 2026-09-15
 chore(launch): declare image input to opencode/codex for vision models

### LINKED ISSUE #509: Declare image input for opencode / codex / claude from the model's capabilities
https://github.com/llmmanorg/llmman/pull/486 made `llmman launch dsh` declare image input when /api/show reports the vision capability. The other integrations don't, so they either never send images to a local vision model or send them to one that can't take them.

HassanBahati:
When `/api/show` reports the `vision` capability for the local model, the opencode config llmman passes in `OPENCODE_CONFIG_CONTENT` now includes image input on the model entry:

​```json
"modalities": { "input": ["text", "image"], "output": ["text"] },
"attachment": true
​```

## Why

opencode assumes a model it doesn't know from models.dev is text-only, so it never attached images to a local vision model. `launch::run` already reads `vision`; before this PR only dsh used it.

For text-only models and `--provider` models, both keys are left out, so their config is byte-identical to before. Claiming image input for a model that can't take it would make opencode send requests the daemon rejects.

## Testing

- New unit test `opencode_config_declares_image_input_only_for_a_vision_model`
- Manual:

gemma4:12b — attached a JPEG image; the model described it correctly. 

image

smollm2:135M-Q4_K_M (text-only) — opencode didn't send the image and replaced it with "this model does not support image input" (checked with `llmman log`).

image

Follow-on to #486, Closes #509


HassanBahati (commented):
Claude Code seems to have no way to declare a model's image input: its `*_SUPPORTED_CAPABILITIES` variables covers only effort and thinking, and apply to Bedrock/Vertex/Foundry rather than a custom `ANTHROPIC_BASE_URL`. It always sends `image` blocks, which `/v1/messages` passes to the local model

HassanBahati:
@codex review

ericcurtin (approved):
LGTM. Matches the upstream shape for both opencode (`modalities`) and the codex catalog entry.

Follow-ons:
- once `/api/show` exposes `<arch>.context_length` (#512), use it for codex `context_window` before falling back to `LLMMAN_CONTEXT_LENGTH`/128000
- same image-input declaration for the other integrations that take a model config (e.g. qwen, goose)

PR #511, opened 2026-09-16
Drop the duplicated HF model info and file list context

ricky-chaoju:
`fetch_model_info` and `fetch_files` each attach their own context, naming the operation at the layer that knows the URL. Six call sites in `hf/pull.rs` and `hf/transfer.rs` attached it a second time, so a failed pull read `HF model info: HF model info: GET …` and a failed file list read `HF file list: HF file list: GET …`.

The four plain callers drop theirs. The three text-encoder callers keep the part that says which encoder failed and lose the repeated noun, leaving `text encoder <ref>: HF model info: GET …`.

Verified by driving `hf::pull` with `HF_ENDPOINT` pointed at a local endpoint, once unroutable and once a stub returning a minimal model info and a 404 for the tree, so both labels are reached without a network. Each went from two occurrences on main to one on this branch.


ricky-chaoju:
@codex review

ericcurtin (approved):
LGTM. One context per layer, all six call sites covered.

ericcurtin:
Merged, thanks. Follow-ups:
- Audit the other `.context(...)` callers of `hf::api` (e.g. `get_json`/download paths) for the same double-labelling.
- Add a test asserting the error chain for a failed `fetch_files` has exactly one `HF file list`, so this cannot regress silently.

PR #504, opened 2026-09-15
Move the wire-refusal helpers into serve/refusal.rs

ricky-chaoju:
Moves the three wire-refusal helpers out of `serve.rs` into a new `serve/refusal.rs`, the module the #498 merge note asked for by name. `unsupported_on_wire` and `wire_refusal` refuse a route up front from what the provider's wire is known to carry; `explain_missing_route` rewrites a 404 that already came back. #498 left the three contiguous under their own banner, so the cut is a single run and the banner gives way to the module doc.

`serve.rs` calls none of them, so nothing there imports them back: `openai.rs`, `ollama.rs` and `serve/tests.rs` reach the new module directly, and all three become `pub(super)`. Those three keywords are the only edits in transit, leaving 61 of the 64 moved lines byte-identical. No tests move and no test body changes.

Related https://github.com/llmmanorg/llmman/issues/429

PR #518, opened 2026-09-17
Move the proxy and relay helpers into serve/relay.rs

ricky-chaoju:
PR #504 stopped one section short of the proxy code below the refusal helpers; ericcurtin's review named the rest, asking that `convert_upstream` and the relay helpers follow into `serve/relay.rs`. The new module takes the contiguous proxy/relay run, `relay_chat_upstream`, and the SSE conversion trait with its macro and one invocation, which is textually scoped and cannot be separated from `convert_upstream`. `serve.rs` falls from 4478 lines to 4138.

The boundary was the question. `relay_chat_upstream` sits outside that run and takes a `ChatUpstream`, but leaving it behind would silently retarget its own doc link on `relay` at the new module of the same name, and a descendant already sees an ancestor's private type, so it travels with no visibility change in `serve.rs`. `SseConverter` had to become `pub(super)`, or `private_bounds` fires on the moved `convert_upstream`. Otherwise the moved text is verbatim, in three commits by group!!!


ricky-chaoju:
@codex review

ericcurtin (approved):
Verified with `--color-moved`: verbatim move, only visibility and import changes. Clippy, fmt and serve tests pass locally.

ericcurtin:
Merged, thanks! Follow-ons if you're keen:
- `serve.rs` is still ~4.1k lines: the model-load/`ensure_model`/`request_pin` block and the hybrid-fallback code (`send_with_hybrid_fallback`, `provider_compat`) are the next self-contained runs.
- `relay.rs` has no tests of its own; the `rewrite_*_model` unit tests in `serve/tests.rs` could move alongside.

PR #519, opened 2026-09-17
feat: forward Codex and Claude OAuth requests over authenticated TLS

ankur-docker:
## Summary

- add an authenticated loopback TLS listener for managed inference
- forward Codex and Claude OAuth requests without persisting credentials
- validate profiles, upstream addresses, headers, and model namespaces
- add documentation and comprehensive tests

## Testing

- `cargo fmt --check`
- `NO_PROXY=127.0.0.1,localhost,::1 no_proxy=127.0.0.1,localhost,::1 cargo test`
- Core and CLI suites passed; optional launch suite passed 15/16, with Qwen skipped by the environment because it is not installed.


src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,723 @@
...
+                || (a == 169 && b == 254)
+                || (a == 172 && (16..=31).contains(&b))
+                || (a == 192 && ((b == 0 && (c == 0 || c == 2)) || b == 168))
+                || (a == 198 && (b == 18 || b == 19 || (b == 51 && c == 100)))
+                || (a == 203 && b == 0 && c == 113))
+        }
+        IpAddr::V6(ip) => {
+            let words = ip.segments();
+            words[0] & 0xe000 == 0x2000
+                && !(words[0] == 0x2001 && (words[1] < 0x0200 || words[1] == 0x0db8))
```
ricky-chaoju, line 514: Nit: `2002::/16` passes this. 6to4 embeds an IPv4, so `2002:7f00:0001::` is 127.0.0.1 and `2002:0a00:0001::` is 10.0.0.1 — both come back `true` from `public_upstream_ip`. Same trick as the `::ffff:127.0.0.1` you already reject. `&& words[0] != 0x2002` covers it.

ricky-chaoju (commented):
LGTM.

src/cmd/serve/managed.rs
ankur-docker: comment is addressed. Please check. Thanks!

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+    "x-ratelimit-reset-requests",
+    "x-ratelimit-limit-tokens",
+    "x-ratelimit-remaining-tokens",
+    "x-ratelimit-reset-tokens",
+];
+
+#[derive(Deserialize)]
+#[cfg_attr(test, derive(Clone))]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub(super) struct Config {
```
ericcurtin: Move to `src/config.rs` as `ManagedConf` under `[managed]`, string fields, `#[serde(default)]`, matching `AuthConf`. Then `llmman config set managed.x y` just works.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+    listen: SocketAddr,
+    tls_cert_file: PathBuf,
+    tls_key_file: PathBuf,
+    connection_key_file: PathBuf,
+    ready_file: PathBuf,
+}
+
+impl Config {
+    pub(super) fn from_env() -> Result<Option<Self>> {
+        let Some(path) = std::env::var_os("LLMMAN_MANAGED_CONFIG") else {
```
ericcurtin: Drop the env var and the JSON format; read from `crate::config::files()` like `auth_api_keys()` does.

ericcurtin (commented):
Thanks for this, a lot of it is really well done:

- Credential handling is right: per-request bearer, never persisted, never falls back to stored keys, redacted from errors.
- Redirects refused, proxies/env not inherited, duplicate JSON keys rejected, `anthropic-beta` merged instead of clobbered.
- Streaming with backpressure and cancel-on-disconnect.
- Tests are broad and generate TLS fixtures at runtime instead of committing a key.

One design ask before this goes further: **no second config file, and no second listener.**

Config: `~/.config/llmman/llmman.conf` already exists, is TOML, has `deny_unknown_fields` sections, and `llmman config get|set|list` works on any dotted key with zero CLI changes. Add a `[managed]` table there (`src/config.rs`), reuse `config::owner_readable_only` for the mode gate, resolve relative paths against `File.dir` like `verify.trust[].keys`. That removes `LLMMAN_MANAGED_CONFIG`, `schemaVersion`, `read_private_file`, and the JSON docs.

```toml
[managed]
enabled = "true"
```

Listener: mount the forwarding routes on the existing daemon under a prefix (e.g. `/api/codex/*path`, `/api/anthropic/*path`), and enforce loopback by checking the peer address of the connection, not by opening a second socket. TLS and API keys then come from `LLMMAN_TLS_CERT/KEY` and `[auth] api_keys`, which already exist. Readiness is `GET /api/version`, which already exists. That deletes the ready file, `instanceId`, `connectionKeyFile`, `Listener`, `Drop`, `write_atomic_with_mode`, the `join!` change in `serve.rs`, the `auth.rs` special case, and all three Copilot findings at once.

Upstream selection: pick the upstream from what the request already carries (`ChatGPT-Account-ID` present -> chatgpt.com; `/api/anthropic` path -> api.anthropic.com) instead of a custom `X-LLMMan-Auth-Profile` header. Resolve DNS in the proxy; drop `X-LLMMan-Upstream-IP`.

Headers: pass native requests through, strip only hop-by-hop headers plus `Host` and our own internal ones. A hard-coded allowlist of provider headers will break the first time a provider adds one.

Smaller items inline. I'm moving this to draft so the redesign can land in peace; please mark ready for review again when it is, and I'll re-review promptly.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+            &self.connection_key_file,
+            &self.ready_file,
+        ] {
+            anyhow::ensure!(path.is_absolute(), "managed file paths must be absolute");
+        }
+        Ok(())
+    }
+}
+
+fn read_private_file(path: &Path) -> Result<Vec<u8>> {
```
ericcurtin: Duplicates `config::owner_readable_only` (`src/config.rs:758`). Reuse it.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+    socket: Option<std::net::TcpListener>,
+    tls: axum_server::tls_rustls::RustlsConfig,
+    state: ManagedState,
+    ready_file: PathBuf,
+}
+
+impl Listener {
+    /// Requires the process-level rustls provider to be installed already;
+    /// `serve_async` does that once before any listener is built.
+    pub(super) async fn bind(config: Config) -> Result<Self> {
```
ericcurtin: Suggest no second socket: register these routes on the main router and reject non-loopback peers with `ConnectInfo<SocketAddr>`. TLS/keys/readiness then come for free from the existing daemon.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+        )
+        .handle(handle)
+        .serve(router(self.state.clone()).into_make_service())
+        .await;
+        shutdown.abort();
+        result.context("serve managed inference")
+    }
+}
+
+impl Drop for Listener {
```
ericcurtin: Read-then-delete is racy with a successor (Copilot flagged this). Goes away if readiness is `GET /api/version` instead of a file.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+fn ready_file_is_ours(path: &Path, ready: &Ready) -> bool {
+    std::fs::read(path)
+        .ok()
+        .and_then(|bytes| serde_json::from_slice::<ReadyIdentity>(&bytes).ok())
+        .is_some_and(|published| {
+            published.instance_id == ready.instance_id && published.pid == ready.pid
+        })
+}
+
+fn write_ready(path: &Path, ready: &Ready) -> Result<()> {
```
ericcurtin: `ready_file` may equal `tls_key_file` and overwrite it. Also goes away without the ready file.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+    fn upstream_client(&self, profile: Profile, ip: IpAddr) -> Result<reqwest::Client> {
+        let mut clients = self.clients.lock().unwrap_or_else(|e| e.into_inner());
+        if let Some(client) = clients.get(&(profile, ip)) {
+            return Ok(client.clone());
+        }
+        // Do not inherit HTTP(S)_PROXY/ALL_PROXY or LLMMAN_TLS_CA. This profile
+        // promises its supervisor exactly this validated HTTPS destination.
+        let client = reqwest::Client::builder()
+            .no_proxy()
+            .resolve(profile.host(), SocketAddr::new(ip, 443))
```
ericcurtin: Caller-supplied IP pinning is unusual and pushes DNS into every client. Resolve here; let the supervisor enforce egress policy at the network layer.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+
+impl Credentials {
+    fn parse(headers: &HeaderMap, profile: Profile) -> Result<Self, AppError> {
+        if single_header(headers, AUTH_PROFILE)? != Some(profile.name()) {
+            return Err(error(
+                StatusCode::BAD_REQUEST,
+                "managed auth profile does not match the requested operation",
+            ));
+        }
+        let upstream_ip = single_header(headers, UPSTREAM_IP)?
```
ericcurtin, line 235: Drop `X-LLMMan-Upstream-IP` and `public_upstream_ip` entirely (see above).

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+const CAPABILITY: &str = "codex-oauth-forwarding-v1";
+const CLAUDE_CAPABILITY: &str = "claude-oauth-forwarding-v1";
+const CLAUDE_OAUTH_BETA: &str = "oauth-2025-04-20";
+const BODY_LIMIT: usize = 32 * 1024 * 1024;
+const ERROR_LIMIT: usize = 1024 * 1024;
+/// Upstream transports kept warm, one per (profile, pinned IP). The
+/// supervisor pins a handful of resolved addresses, so this is a leak
+/// guard rather than an eviction policy: past it the map starts over.
+const CLIENT_CACHE_LIMIT: usize = 64;
+const CODEX_REQUEST_HEADERS: &[&str] = &[
```
ericcurtin: Prefer a hop-by-hop denylist (`connection`, `keep-alive`, `transfer-encoding`, `te`, `trailer`, `upgrade`, `proxy-*`) plus `host` and our internal headers, forward everything else unchanged. Same for `RESPONSE_HEADERS`.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+    }
+    for name in profile.request_headers() {
+        if let Some(value) = single_header(&headers, name)? {
+            request = request.header(*name, value);
+        }
+    }
+    match profile {
+        Profile::Codex => {
+            if !headers.contains_key("originator") {
+                request = request.header("originator", "codex_cli_rs");
```
ericcurtin: Docs say llmman does not invent client identity, but this injects `originator: codex_cli_rs` and `openai-beta`. Forward what the caller sent; do not default these.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,724 @@
...
+                beta.to_owned()
+            } else if beta.trim().is_empty() {
+                CLAUDE_OAUTH_BETA.to_owned()
+            } else {
+                format!("{beta},{CLAUDE_OAUTH_BETA}")
+            };
+            request = request.header("anthropic-beta", beta);
+        }
+    }
+    let mut upstream = request.send().await.map_err(|_| {
```
ericcurtin: `map_err(|_| ...)` discards the reqwest error. Log it (redacted) at debug so a failed upstream connect is diagnosable.

src/cmd/serve/managed/tests.rs
```diff
@@ -0,0 +1,1008 @@
...
+        connection_auth: auth::Policy::with_keys(["connection-secret"]),
+        ready: Ready {
+            schema_version: 1,
+            instance_id: "instance-123".into(),
+            pid: std::process::id(),
+            endpoint: "https://127.0.0.1:45678".into(),
+            capabilities: vec![CAPABILITY, CLAUDE_CAPABILITY],
+        },
+        clients: std::sync::Mutex::default(),
+        test_base: Some(base.into()),
```
ericcurtin, line 36: `test_base` bypasses `.resolve(profile.host(), ...)`, so the pinning is never exercised. Either test through the real client builder or drop pinning (preferred).

src/cmd/serve.rs
```diff
@@ -4286,6 +4288,11 @@ async fn serve_async(_args: &ServeArgs) -> anyhow::Result<()> {
         );
     }
     let tls = tls_from_env()?;
+    let managed_config = managed::Config::from_env()?;
```
ericcurtin: Should read `crate::config::managed()` (new accessor), not an env var.

src/cmd/serve.rs
```diff
@@ -4286,6 +4288,11 @@ async fn serve_async(_args: &ServeArgs) -> anyhow::Result<()> {
...
+    // Both rustls providers are compiled in (reqwest's, the AWS SDK's), so
+    // none is the default until one is installed. Once, here, before either
+    // listener builds a TLS config.
+    let _ = rustls::crypto::ring::default_provider().install_default();
```
ericcurtin: Revert to only installing under `Some((cert, key))` unless a second TLS listener remains.

src/cmd/serve.rs
```diff
@@ -4398,39 +4412,59 @@ async fn serve_async(_args: &ServeArgs) -> anyhow::Result<()> {
...
-                .await?
+        Ok::<(), anyhow::Error>(())
+    };
+    // `join`, not `try_join`: a failure on the managed listener must not
+    // drop the public server mid-request or skip the unload below. Its
+    // error is logged as it happens (its ready file is retracted as its
+    // listener drops) and reported once the public server has drained.
+    let (public, managed) = match managed {
+        Some(managed) => {
+            tokio::join!(public_server, async {
```
ericcurtin: If `public_server` errors, `managed` keeps running until the shutdown signal (Copilot flagged). Not needed if routes share the one listener.

src/cmd/serve/auth.rs
```diff
@@ -130,6 +130,17 @@ pub(super) async fn require_key(
     mut req: Request,
     next: Next,
 ) -> Response {
+    // Managed provider credentials are accepted only on the separate TLS
+    // listener. Never reinterpret that profile as ordinary API-key traffic.
+    if req.headers().contains_key(super::managed::AUTH_PROFILE) {
```
ericcurtin: Not needed with a single listener; the `/api/codex` routes just sit behind the same `require_key`.

src/fsutil.rs
```diff
@@ -16,6 +16,18 @@ static TMP_COUNTER: AtomicU64 = AtomicU64::new(0);
...
 /// the temp is synced before the rename.
 pub fn write_atomic(path: &Path, bytes: &[u8]) -> std::io::Result<()> {
+    write_atomic_inner(path, bytes, None)
+}
+
+/// [`write_atomic`], but the file ends up with exactly `mode` (Unix)
+/// whether or not it already existed, for documents that must stay
+/// owner-only regardless of the umask or what a previous writer left.
+/// Elsewhere the mode is ignored.
+pub fn write_atomic_with_mode(path: &Path, bytes: &[u8], mode: u32) -> std::io::Result<()> {
```
ericcurtin: Nice helper, but unused once the ready file goes. Drop it from this PR.

Cargo.toml
```diff
@@ -120,6 +120,9 @@ fuzzing = []
 # A WebSocket client for the `/llmman/shell` tests; the same crate axum's
 # `ws` feature already builds, at the version it pins.
 tokio-tungstenite = "0.24"
+# Managed-listener tests create short-lived TLS identities instead of storing
+# a private key in the repository.
+rcgen = "0.14"
```
ericcurtin, line 128: Not needed if tests run over plain HTTP on the shared router.

docs/providers.md
```diff
@@ -272,3 +272,116 @@ block and user block), a `response_format` JSON schema becomes a forced
...
+
+## OAuth credential forwarding
+
+A caller can supply a current Codex or Claude OAuth access token to llmman
+for one native request. The caller owns login, token refresh, account
+selection, authorization, and network policy.
+llmman does not read or persist a refresh token. Ordinary provider API-key
+routing and the public listener keep their existing behavior.
+
+Set `LLMMAN_MANAGED_CONFIG` to an owner-only JSON file to enable an additional
```
ericcurtin: Document the `[managed]` table in `docs/configuration.md` alongside `[auth]` instead; no env var, no JSON.

docs/providers.md
```diff
@@ -272,3 +272,116 @@ block and user block), a `response_format` JSON schema becomes a forced
...
+| `codex-oauth` | `POST /v1/responses/compact` | `https://chatgpt.com/backend-api/codex/responses/compact` |
+| `claude-oauth` | `POST /v1/messages` | `https://api.anthropic.com/v1/messages` |
+| `claude-oauth` | `POST /v1/messages/count_tokens` | `https://api.anthropic.com/v1/messages/count_tokens` |
+
+Codex requests require these headers, in addition to the connection key:
+
+```text
+Authorization: Bearer <current access token>
+X-LLMMan-Auth-Profile: codex-oauth
+X-LLMMan-Upstream-IP: <policy-authorized public IP of chatgpt.com>
```
ericcurtin: Drop with the header.

docs/providers.md
```diff
@@ -272,3 +272,116 @@ block and user block), a `response_format` JSON schema becomes a forced
...
+X-LLMMan-Upstream-IP: <policy-authorized public IP of api.anthropic.com>
+```
+
+The JSON `model` must be `llmman.provider/anthropic/<exact-model-id>`.
+llmman sends the supplied token as a bearer credential, never as `x-api-key`.
+It adds `oauth-2025-04-20` to `anthropic-beta` when absent, preserving other
+beta features, and defaults `anthropic-version` to `2023-06-01` when omitted.
+The caller supplies the native Messages payload, including any client-specific
+system content or metadata required by the provider; llmman does not invent
+client identity or alter prompts.
```
ericcurtin: Contradicts the `originator`/`openai-beta` defaults in `forward`.

docs/providers.md
```diff
@@ -272,3 +272,84 @@ block and user block), a `response_format` JSON schema becomes a forced
...
+
+## OAuth credential forwarding
+
+Enable `[managed]` in [configuration.md](configuration.md#managed-oauth-forwarding)
+to forward native Codex and Claude requests on the daemon's existing TLS
+listener. The caller owns login, refresh, account selection, authorization,
+and network policy. llmman never persists these request-local OAuth tokens or
+falls back to stored provider keys, environment credentials, or peers.
+
+Every forwarding request must come from a loopback connection and supply both:
```
ericcurtin: "both" but three headers follow. "all of".

docs/providers.md
```diff
@@ -272,3 +272,84 @@ block and user block), a `response_format` JSON schema becomes a forced
...
+forwarded only when supplied; llmman does not invent a client identity.
+For Claude, llmman merges `oauth-2025-04-20` into `anthropic-beta` and defaults
+`anthropic-version` to `2023-06-01` when absent. The caller supplies any required
+native system content or metadata.
+
+Successful response bodies stream with backpressure and cancellation on
+disconnect. Upstream connection establishment has a 30-second timeout; reads
+have no deadline unless `managed.read_timeout_seconds` is configured. That
+optional timeout applies to response-header waits and individual stalled reads,
+not total generation duration. A quiet generation can exceed it, so choose a
```
ericcurtin: Backwards: a quiet generation is exactly what an idle-read timeout cuts off; an active stream can outlive it. Flip the sentence.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,597 @@
...
+async fn require_connection(
+    State(state): State<ManagedState>,
+    mut req: Request,
+    next: Next,
+) -> Response {
+    // Only connection metadata is authoritative; forwarded IP headers are not.
+    let local = req
+        .extensions()
+        .get::<ConnectInfo<SocketAddr>>()
+        .is_some_and(|peer| peer.0.ip().is_loopback());
```
ericcurtin: `ip().to_canonical().is_loopback()` — on a dual-stack bind a local IPv4 peer arrives as `::ffff:127.0.0.1` and is refused here.

src/cmd/serve/managed.rs
```diff
@@ -0,0 +1,597 @@
...
+                || (a == 100 && (64..=127).contains(&b))
+                || (a == 169 && b == 254)
+                || (a == 172 && (16..=31).contains(&b))
+                || (a == 192 && ((b == 0 && (c == 0 || c == 2)) || b == 168))
+                || (a == 198 && (b == 18 || b == 19 || (b == 51 && c == 100)))
+                || (a == 203 && b == 0 && c == 113))
+        }
+        IpAddr::V6(ip) => {
+            let words = ip.segments();
+            words[0] & 0xe000 == 0x2000
```
ericcurtin, line 354: Still prefer resolving here: hostname verification already bounds where the bearer lands, egress is the supervisor's network layer. Not blocking if Sandboxes needs it today, but please file an issue to drop `X-LLMMan-Upstream-IP`.

Nit while it stays: `3fff::/20` (RFC 9637 doc prefix) passes.

src/cmd/serve.rs
```diff
@@ -3777,6 +3778,15 @@ async fn serve_async(_args: &ServeArgs) -> anyhow::Result<()> {
         );
     }
     let tls = tls_from_env()?;
+    let managed_config = crate::config::managed().map_err(anyhow::Error::msg)?;
+    if managed_config.enabled {
+        if let Some((_, key)) = &tls {
+            crate::config::owner_readable_only(key)
```
ericcurtin, line 3784: `owner_readable_only` is `Ok(())` off unix, so the documented owner-only requirement is not enforced on Windows. Either say so in the docs or refuse managed mode there.

ericcurtin (changes requested):
Much closer to the shape we discussed, thanks. A few small things inline, plus: the PR description still says "This PR remains draft" and that Sandboxes needs a coordinated migration — please update it. Setting to draft; mark ready when done and I'll re-review.

ericcurtin (commented):
Thanks, all of the last round is addressed. Two small things:
- The test build breaks after rebasing on main. #538 moved `test_state` into `serve/test_support.rs`, so `managed/tests.rs:1088` should use `super::super::test_support::test_state()`.
- Please file the follow-up issue to drop `X-LLMMan-Upstream-IP`.

Moving to draft for now. Please mark it ready for review again after rebasing.

ericcurtin (approved):
LGTM.

ericcurtin:
Merged, thanks. Follow-ons: #543; drop the "earlier draft" migration paragraph from docs/providers.md (never shipped); make `Conf.managed` private behind `managed()` like `auth`.

PR #521, opened 2026-09-18
chore: offer to install opencode and qwen when missing

HassanBahati:
When `opencode` or `qwen` isn't installed, `llmman launch` now asks to install it with the official installer, then launches it.  

Follow-on to #486


HassanBahati:
@ericcurtin claude, codex and gemini seem to have bigger downloads...should we as well have them installed with a npx execution incase they aren't installed ?

ricky-chaoju:
> @ericcurtin claude, codex and gemini seem to have bigger downloads...should we as well have them installed with a npx execution incase they aren't installed ?

I think size is the wrong cut here, dsh is 295 MB installed, codex 313, claude 222, opencode 177, qwen 138, gemini 100. gemini's the smallest of the lot.

claude npx'd fine for me. codex I'd leave out, ci.yml already has a note about `npm install -g @openai/codex` exiting 0 with a broken binary. gemini's not in ci.yml at all.
@ericcurtin WDYT?

ericcurtin (changes requested):
Thanks! One design point: npx-per-launch fits dsh, but for opencode and qwen the expected behaviour is a real install, not a one-off run:

- Not found → `OpenCode is not installed. Install now? [y/N]`; decline → error "opencode installation cancelled".
- opencode: linux/macOS `bash -c 'set -o pipefail; curl -fsSL https://opencode.ai/install | bash'`, windows `npm install -g opencode-ai@latest`. Also check `~/.opencode/bin/opencode` when resolving.
- qwen: the official install-qwen.sh (unix, with its trailing `exec qwen` stripped) / install-qwen.bat (windows via powershell); check `~/.npm-global/bin` and `%APPDATA%\npm` locations.
- Check installer deps first (curl+bash, or npm/powershell) and print what's missing; after install re-resolve and fail with "installed but not found on PATH, restart your shell".
- Keep the npx path for dsh only.

Setting to draft — please re-open for review when ready.

HassanBahati:
Performed some manual tests
image

image

ericcurtin (approved):
Faithful port of the reference install flow (deps check, prompt, installer command, qwen shim, post-install lookup). Merging.

Follow-ons:
- Create the shim dir with `create_dir` (fail if it exists) rather than `create_dir_all`, or a random name, so a pre-existing dir can't be reused.
- Consider a `--yes` flag to skip the prompt (non-interactive/CI).
- Same install-on-demand for codex/claude/gemini.

PR #522, opened 2026-09-18
chore: use model's context for codex's context_window

HassanBahati:
In https://github.com/llmmanorg/llmman/pull/512,  `/api/show` exposed `<arch>.context_length` , this PR aims to use it for codex context_window before falling back to LLMMAN_CONTEXT_LENGTH/128000

Related https://github.com/llmmanorg/llmman/pull/510


HassanBahati:
Something interesting I came across:

When `LLMMAN_CONTEXT_LENGTH` is unset, the daemon serves the model's `context_length`, say 131072, because that's what it defaults to when the variable isn't set. Codex uses the same 131072 in that scenario, since it's the model's `context_length`, so the two stay in sync.

An edge case comes up when `LLMMAN_CONTEXT_LENGTH` is set, for instance to 16384. The daemon serves 16384, because an explicit value always takes priority over the model's `context_length`. Codex, however, still uses 131072, because it takes the model's `context_length` first and only falls back to `LLMMAN_CONTEXT_LENGTH` when there isn't one.

| | `LLMMAN_CONTEXT_LENGTH` unset | `LLMMAN_CONTEXT_LENGTH=16384` |
|---|---|---|
| Daemon serves | 131072 | 16384 |
| Codex uses | 131072 | 131072 |

In that case codex thinks it has more room than it does and waits too long to compact, so long sessions can fail with a "context size exceeded" error. Shorter sessions aren't affected, since they still fit within 16384.

Curious to hear your thoughts, @ericcurtin @ricky-chaoju.

ricky-chaoju:
@HassanBahati I'd flip the precedence. `context_window` is what codex compacts against, so it needs to be whatever the daemon actually takes. On main the explicit case already works; this PR trades it away for the unset one.

Move `.or(trained)` after the env parse: unset still picks up the trained context, 16384 keeps working. Two rows of the new test would flip.

Longer term `/api/ps` already reports the live `n_ctx` from llama-server's `/props`, which is the only number that survives the OOM auto-shrink. Probably a follow-up though. WDYT?

HassanBahati:
@ricky-chaoju I agree. Flipping it keeps the explicit case working and still picks up the trained context when unset.

src/cmd/launch.rs
```diff
@@ -906,22 +923,26 @@ fn write_codex_file(path: &Path, contents: &str) -> anyhow::Result<()> {
...
-        .ok()
-        .and_then(|v| v.trim().parse::<u64>().ok())
-        .filter(|n| *n > 0)
+/// The model's `trained` context, else a positive `LLMMAN_CONTEXT_LENGTH`
+/// (`env`), else [`CODEX_FALLBACK_CONTEXT_WINDOW`]. An `env` below the
+/// trained context is what the daemon serves, so codex is told more.
+fn codex_context_window(env: Option<&str>, trained: Option<u64>) -> u64 {
+    trained
+        .or_else(|| {
+            env.and_then(|v| v.trim().parse::<u64>().ok())
```
ericcurtin: Precedence is inverted. `LLMMAN_CONTEXT_LENGTH` is the `--ctx-size` the server actually runs with, so when set (>0) it must win; trained context is the fallback, then 128000. Telling codex a window larger than what's served means it overflows before compacting.

ericcurtin (changes requested):
Reading the header via `/api/show` is the right approach. One precedence issue to fix; setting to draft, please mark ready when done and I'll re-review. #521 just merged into `launch.rs`, so rebase on main too.

HassanBahati:
Resolved the precedence; `LLMMAN_CONTEXT_LENGTH` (> 0) → trained context → 128000. Launch also now reads `LLMMAN_CONTEXT_LENGTH` the same way the daemon does.

ericcurtin (approved):
Precedence matches what the daemon serves now; merges cleanly on main, clippy clean, tests pass.

ericcurtin:
Merged, thanks. Follow-ons:
- opencode's `limit.context` and the other launchers that declare a window (kimi, deepseek harness) can reuse `ShowResponse::context_length()` the same way.
- Hybrid pairs: the hosted half's window differs from the local one; worth deciding which the catalog should declare.

PR #523, opened 2026-09-19
Reuse blob links for model cache files

LxYuan0420:
This PR is:

- To follow up on #460.
- To use the same hardlink-or-copy path for named blob cache files and safetensors cache files.
- To try hardlinks on every platform before falling back to atomic copy.
- To make cache pruning account for hardlinked files by file identity.
- To remove stale cache copy `.tmp` files during prune.

#460 stopped duplicating safetensors weights when hardlinks can be used. This follow-up applies the same behavior to named blob cache files and keeps prune reporting correct when blob/cache paths point to the same underlying file.


src/storage/gc.rs
```diff
@@ -172,36 +251,56 @@ fn is_older_than(path: &Path, grace: Duration) -> bool {
...
+            stats.count += inner.count;
+            stats.bytes += inner.bytes;
+            continue;
+        }
+        let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+            continue;
+        };
+        if !name.ends_with(".tmp") || !is_older_than(&path, grace) {
+            continue;
+        }
```
ericcurtin, line 392: Too broad: any layer file named `*.tmp` in a live cache dir gets deleted (and hardlinks inherit the blob's old mtime, so the grace period won't save it). Match the exact scratch suffix `.<pid>.<counter>.tmp` that `cache_copy_temp_path` emits.

src/storage/gc.rs
```diff
@@ -94,12 +156,13 @@ pub fn prune_blobs(
...
         }
         if live.contains(&format!("sha256:{name}")) {
+            account.remember(&path);
             continue;
         }
         if !is_older_than(&path, grace) {
             continue;
         }
-        let size = freed_file_size(&path);
+        let size = account.size(&path);
```
ericcurtin, line 265: Nit: blobs kept only by the grace period aren't `remember`ed, so deleting their cache link reports the full size. Remember them too so bytes aren't overstated.

ericcurtin (changes requested):
Looks good overall. One thing to fix before merge; setting to draft, please mark ready when done and I'll re-review.

src/storage/gc.rs
```diff
@@ -40,6 +40,45 @@ pub struct GcStats {
     pub bytes: u64,
 }
 
+#[derive(Default)]
+struct GcFileAccount {
+    files: HashSet<same_file::Handle>,
```
ericcurtin: `same_file::Handle` keeps the `File` open for its lifetime, so this holds one fd per live blob and per cache file for the whole sweep. macOS's default soft limit is 256; past that `remember` silently fails and bytes are overcounted again. On unix key by `(dev, ino)` from `symlink_metadata()` via `MetadataExt` (no fd); keep `same_file` for `cfg(windows)` only.

src/modelpack.rs
```diff
@@ -747,29 +729,39 @@ fn extract_safetensors_dir(
...
-            Err(hardlink_error) => copy_file_atomic(src, dest, layer_size).with_context(|| {
-                format!(
-                    "hardlink {} to {} failed: {hardlink_error}; copy failed",
-                    src.display(),
-                    dest.display()
-                )
-            }),
+    match std::fs::hard_link(src, dest) {
+        Ok(()) if cached_layer_file_matches(dest, layer_size) => Ok(()),
+        Ok(()) => {
```
ericcurtin, line 753: Hardlink succeeded but the size is wrong means the blob itself is short; copying it just repeats the same failure after streaming a multi-GB file. Remove the link and `bail!` with the size mismatch here instead of falling into `copy_file_atomic`.

src/storage/gc.rs
```diff
@@ -75,6 +114,110 @@ pub fn prune_blobs(
     store_root: &Path,
     live: &HashSet<String>,
     grace: Duration,
+) -> anyhow::Result<GcStats> {
+    prune_blobs_with_account(store_root, live, grace, &mut GcFileAccount::default())
```
ericcurtin, line 137: Nit: the startup sweep in `serve.rs` still calls `prune_blobs` + `prune_cache` separately. Switch it to `prune_blobs_and_cache` too, then these two wrappers can become `#[cfg(test)]` or go away.

ericcurtin (changes requested):
Both earlier asks are addressed, thanks. Two things left, then it's good to go. Setting to draft; mark ready when done and I'll re-review.

ericcurtin (approved):
All five asks addressed; verified locally (tests, clippy -D warnings, fmt). Thanks.

Follow-ups worth considering (separate PRs):
- `remember_cache_dir_files` swallows `read_dir`/`file_type` errors, so an unreadable nested dir can still overstate freed bytes; returning `anyhow::Result` and propagating would close that.
- On unix the raw-layer cache now hardlinks instead of symlinking, so a cache dir on a different filesystem than the store gets a full copy; worth a line in the docs for `LLMMAN_CACHE`/`LLMMAN_MODELS`.

ericcurtin:
@LxYuan0420 thanks. Are you interested in becoming a maintainer?

LxYuan0420:
> @LxYuan0420 thanks. Are you interested in becoming a maintainer?

Sounds good! Will contribute as much as I can. Thanks for the invitation!

ericcurtin:
Access granted... @LxYuan0420 and @ricky-chaoju you guys should submit too blog posts individually to 

https://github.com/llmmanorg/llmmanorg.github.io/pulls

that way we can introduce you as a maintainers and show people some nice content. The blog post could be about things you worked on OR things you didn't work on, both work!

PR #524, opened 2026-09-19
launch: stop installing integrations on demand

ericcurtin:
## Summary

`llmman launch opencode` / `llmman launch qwen` prompted to run the integration's official installer when the binary was missing (and printed the install command when there was no terminal to prompt on). How an integration is installed is up to the user; llmman should only exec what is already on the machine.

This removes:

- the `Installer` registry (`installer_for`), the `Install now? [y/N]` prompt (`ensure_installed`), the qwen install shim, and the dependency/install hints
- the associated unit tests
- the README paragraph describing the auto-install behaviour

A missing opencode/qwen is now reported as `<name> is not installed`, the same as every other integration.

Effectively reverts the on-demand install introduced in #521.

## Testing

- `cargo build`, `cargo fmt --check`, `cargo clippy --all-targets`: clean
- `cargo test launch`: 53 unit tests pass; launch e2e tests pass except `launch_claude_with_model`, which fails identically on `main` in this environment (claude exits 1 with no output) and is unrelated to this change


HassanBahati:
@ericcurtin I'm curious what would have been a better approach here. I get that how an integration is installed is up to the user and llmman should only exec what is already on the machine. Should a proactive installation have used tools the user already has instead, like brew, npm or winget, rather than running the official install scripts? Or is any install step out of scope for llmman, so we should just report "not installed" (maybe with a docs link) and leave the rest to the user? Asking so I know where the line is for future implementations.

ericcurtin:
> @ericcurtin I'm curious what would have been a better approach here. I get that how an integration is installed is up to the user and llmman should only exec what is already on the machine. Should a proactive installation have used tools the user already has instead, like brew, npm or winget, rather than running the official install scripts? Or is any install step out of scope for llmman, so we should just report "not installed" (maybe with a docs link) and leave the rest to the user? Asking so I know where the line is for future implementations.

Yeah, I'm on the fence about this tbh... Should we auto-install integrations or leave this to the user? But I think we can still keep moving forward with integrations either way :)

HassanBahati:
Sounds good, thanks! I'll keep going on integrations in the meantime

PR #528, opened 2026-09-19
chore(launch): declare image input to qwen/hermes for vision models

HassanBahati:
Follow-on to #510, which declared image input to opencode and codex. This does the same for the other integrations whose model config llmman writes.

### Changes
- **qwen:** llmman's `modelProviders.openai` entry in `settings.json` gets `generationConfig.modalities.image: true` for a vision model. Qwen Code reads modalities only from the provider entry; a top-level `model.generationConfig` is ignored for provider models. Verified against Qwen Code 0.24.0.
- **hermes:** `config.yaml` gets `model.supports_vision: true` for a vision model. That's the override hermes's image routing reads (`_supports_vision_override`), so images reach the model directly instead of going through its `vision_analyze` tool.

A text-only model gets neither key, so nothing changes for it. If a different `--model` is passed after `--` to qwen, image input isn't declared, since the vision check was done on the model llmman resolved.


HassanBahati:
goose doesn't have a field for declaring image input yet, so it's left out here. I came across https://github.com/aaif-goose/goose/issues/11998, which attempts to add the support, I'll keep an eye on it and follow up once it lands.

src/cmd/launch.rs
```diff
@@ -1310,12 +1321,20 @@ fn launch_openclaw(model: &str, extra_args: &[String]) -> anyhow::Result<()> {
...
+fn launch_qwen(
+    model: &str,
+    api_key: &str,
+    vision: bool,
+    extra_args: &[String],
+) -> anyhow::Result<()> {
     let bin = find_qwen().ok_or_else(|| anyhow::anyhow!("qwen is not installed"))?;
-    let model = forwarded_model(extra_args).unwrap_or(model);
+    // `vision` describes `model`, not a different one forwarded after `--`.
+    let forwarded = forwarded_model(extra_args).filter(|f| *f != model);
```
ricky-chaoju: `model` has been through `resolve_ollama_api` by now, `forwarded_model` gives back whatever was typed. I ran it: `resolve_ollama_api("gemma4:12b")` is `"docker.io/ai/gemma4:12b"`, so

    llmman launch qwen --model gemma4:12b -- --model gemma4:12b

clears `vision` and writes text-only settings for a vision model. Resolving the forwarded value the same way before comparing covers it.

HassanBahati: Good catch, thanks! I was able to reproduce it as well and the entry came out text-only. I have resolved it so that the forwarded value now goes through `resolve_ollama_api` before the comparison, I have also added a test as well covering that exact command.

ricky-chaoju (approved):
LGTM. Thanks

PR #525, opened 2026-09-19
Move the hybrid-pair routing and retry into serve/hybrid.rs

ricky-chaoju:
Moves the hybrid-pair routing and its one retry out of `serve.rs` into a new `serve/hybrid.rs`, the module the #518 review asked for. `resolve_hybrid_side` picks a half and hands the rest of the daemon that half's ordinary reference; `with_hybrid_fallback` sends the request again to the hosted half when the local one refuses it as over its context. `local_context_overflow` and `OVERFLOW_BODY_LIMIT` come along: the retry is their only caller outside tests. `serve.rs` falls from 4138 lines to 3968.

Related https://github.com/llmmanorg/llmman/issues/429

PR #531, opened 2026-09-21
Move the hybrid fallback tests into serve/hybrid.rs

ricky-chaoju:
`hybrid_fallback`, `local_context_overflow` and `OVERFLOW_BODY_LIMIT` were `pub(super)` for one caller each in `serve/tests.rs` and nothing else, which the #525 merge note flagged. Their two tests move into an inline `mod tests` in `hybrid.rs` the shape most of its siblings already have and the three go private in the second commit.


ricky-chaoju:
@codex review

ericcurtin (approved):
Clean refactor: tests colocated with the helpers they exercise, three items go private. CI green.

Follow-ups worth considering (separate PRs):
- `request_pin`, `resolve_hybrid_side`, `with_hybrid_fallback` are still `pub(super)` only for `serve/tests.rs`; their tests could move here the same way.
- `headers_with`, `PAIR`, `HOSTED` in `serve/tests.rs` are now shared fixtures; a small `test_support` module would remove the `super::super::tests` reach-through.

